PR #127, opened 2020-07-27
Add scGen to data integration benchmark

LuckyMD:
This PR is a work in progress to add scGen to the benchmark. This should include:
- [x] a `runScGen()` function 
- [x] adding scGen computation to the runMethods.py
- [x] enabling passing of cell type labels to scGen via the Snakefile
- [x] a test function or notebook for scGen (optional)
- [x] test scGen on snakemake

The code also still needs to be tested within Snakemake and outside.


LuckyMD:
@mumichae it would be great if you could help with the third point there!

LuckyMD:
@danielStrobl let's either work in this branch also for SCANVI, or later import this into your SCANVI branch.

LuckyMD:
@M0hammadL could you check the function here?

M0hammadL:
Hi @LuckyMD  I have checked it, it seems to be identical as our notebook

danielStrobl:
Can't really judge the new snakemake stuff but the scGen function is looking good

PR #129, opened 2020-07-28
added saucie

danielStrobl:
Added saucie, still needs some testing


scIB/integration.py
```diff
@@ -21,6 +21,7 @@
 import rpy2.robjects as ro
 import anndata2ri
 from scipy.sparse import issparse
+import SAUCIE
```
LuckyMD, line 24: let's only import the methods within the function so that we don't require them all to be installed when scIB is installed.

scIB/integration.py
```diff
@@ -283,6 +287,18 @@ def runBBKNN(adata, batch, hvg=None):
         corrected = bbknn.bbknn(adata, batch_key=batch, neighbors_within_batch=25, copy=True)
     return corrected
 
+
+def runSaucie(adata, batch):
```
LuckyMD, line 291: could you add in a docstring here what tutorial you got the parametrizations from?

scIB/integration.py
```diff
@@ -283,6 +287,18 @@ def runBBKNN(adata, batch, hvg=None):
...
+    saucie = SAUCIE.SAUCIE(adata.X.shape[1])
+    loader = SAUCIE.Loader(adata.X.todense(), labels=adata.obs[batch])
```
LuckyMD, line 293: densifying is necessary? If that is true, it will make things quite a bit slower here unfortunately.

scIB/integration.py
```diff
@@ -283,6 +287,18 @@ def runBBKNN(adata, batch, hvg=None):
...
+    saucie.train(loader, steps=1000)
+    adata.obsm['X_emb'] = saucie.get_embedding(loader)[0]
+    adata.X = saucie.get_reconstruction(loader)[0]
```
LuckyMD, line 296: Is this just reconstructed input data, or batch-removed feature data? If the latter, it's fine. But if it's only a reconstruction, we shouldn't include this as it's not the right thing to test.

LuckyMD:
Also, for the notebook, please check that integration worked by at least showing a umap output before and after.

scIB/integration.py
danielStrobl: Yes, I tested without densifying and it didn't work

danielStrobl:
I have only tested with my small_test.h5ad so far but I'll try with simulations later today

LuckyMD:
Could you add SCANVI in a separate PR? That would make this a lot cleaner I think.

scIB/integration.py
danielStrobl: It's the batch corrected feature data

LuckyMD:
Hi @mattamodio, 
Dan Burkhardt mentioned you might be available to check our SAUCIE batch removal function to see if it is implemented in the way you would expect SAUCIE to be used for the removal of batch effects. We will use the parameter set implemented here to benchmark the method across datasets without further hyperparameter optimization to see how it works out-of-the-box. It would be great to get your feedback on this.

scIB/integration.py
```diff
@@ -283,49 +195,27 @@ def runBBKNN(adata, batch, hvg=None):
...
+    data = pca_op.fit_transform(expr)
+    saucie = SAUCIE.SAUCIE(100, lambda_b=0.1)
+    loader_train = SAUCIE.Loader(data, labels=adata.obs[batch].cat.codes, shuffle=True)
+    loader_eval = SAUCIE.Loader(data, labels=adata.obs[batch].cat.codes, shuffle=False)
+    saucie.train(loader_train, steps=5000)
+    ret = adata.copy()
+    ret.obsm['X_emb'] = saucie.get_reconstruction(loader_eval)[0]
+    ret.X = pca_op.inverse_transform(ret.obsm['X_emb'])
+                                                    
+    return ret
```
LuckyMD, line 216: @mattamodio Here is the `runSaucie()` function that we would be testing.

scripts/runIntegration.py
```diff
@@ -58,7 +58,9 @@ def runIntegration(inPath, outPath, method, hvg, batch):
         'mnn': scIB.integration.runMNN,
         'bbknn': scIB.integration.runBBKNN,
         'scvi': scIB.integration.runScvi,
-        'combat': scIB.integration.runCombat
+        'combat': scIB.integration.runCombat,
+        'scanvi': scIB.integration.runScanvi,
```
LuckyMD, line 62: you still have `scanvi` in here, @danielStrobl ;)

scIB/integration.py
```diff
@@ -22,6 +22,7 @@
 import anndata2ri
 from scipy.sparse import issparse
 
+
```
LuckyMD, line 25: why the space?

scIB/integration.py
```diff
@@ -283,49 +195,27 @@ def runBBKNN(adata, batch, hvg=None):
...
+    import SAUCIE
+    import sklearn.decomposition
+    pca_op = sklearn.decomposition.PCA(100)
+    expr = adata.X.todense()
+    data = pca_op.fit_transform(expr)
+    saucie = SAUCIE.SAUCIE(100, lambda_b=0.1)
+    loader_train = SAUCIE.Loader(data, labels=adata.obs[batch].cat.codes, shuffle=True)
+    loader_eval = SAUCIE.Loader(data, labels=adata.obs[batch].cat.codes, shuffle=False)
+    saucie.train(loader_train, steps=5000)
+    ret = adata.copy()
```
LuckyMD, line 212: quick question, why copy and not just add to the initial adata?

scIB/integration.py
danielStrobl: no idea where that came from...

scIB/integration.py
danielStrobl: I didn't want to overwrite adata.X

LuckyMD: good!

PR #131, opened 2020-07-29
[WIP] DESC integration

LuckyMD:
This PR adds DESC into the benchmarking setup. 

This method requires the following packages to be installed:
`pip install desc==2.0.3 pydot`
`conda install GraphViz`

This is still work in progress. These steps have been taken:
- [x] Add DESC function to scIB
- [x] Test DESC function in notebook
- [x] Test DESC in the Snakemake setup
- [x] Write a test for DESC (optional)


LuckyMD:
@eleozzr Could you check if this function matches how you would use DESC?

LuckyMD:
I think this is about ready, @danielStrobl can you review in case there's sth you can find?

scIB/integration.py
```diff
@@ -324,22 +324,37 @@ def runConos(adata, batch, hvg=None):
...
+    sc.pp.combat(adata, key=batch)
+    return adata
 
 
-if __name__=="__main__":
-    adata = sc.read('testing.h5ad')
-    #emb, corrected = runScanorama(adata, 'method', False)
-    #print(emb)
-    #print(corrected)
+def runDESC(adata, batch, res=0.8, ncores=24, tmp_dir='/localscratch/'):
```
danielStrobl, line 332: Maybe add the source for the parameters to the docstring. Looks good otherwise

danielStrobl, line 332: Is it necessary to set 24 cores manually?

LuckyMD: It will otherwise use half the ones available... So I think we should hardcode the full amount we have, and then let it use half available when there are less than that on the system. Does that make sense to you?

LuckyMD: good point!

danielStrobl: ok, sounds fair. We could also do `os.cpu_count()`

LuckyMD: 883ef0e

LuckyMD: that does sound better! Let's hope that works on the server as well... would need to test again I guess.

LuckyMD: c2eb7e2

LuckyMD:
To do:
- [ ] final snakemake test

PR #132, opened 2020-07-29
add scanvi to scib

danielStrobl:
separate PR for scanVI


LuckyMD:
@romain-lopez it would be great if you could take a look at the function here to see if this is the parametrization you would expect for SCANVI.

scIB/integration.py
```diff
@@ -174,6 +174,69 @@ def runScvi(adata, batch, hvg=None):
...
+    del net_adata.raw # Note that this only works from anndata 0.7
+
+    # Define batch indices
+    le = LabelEncoder()
+    net_adata.obs['batch_indices'] = le.fit_transform(net_adata.obs[batch].values)
+    net_adata.obs['labels'] = le.fit_transform(net_adata.obs[labels].values)
+
+    net_adata_out = AnnDatasetFromAnnData(net_adata)
+    net_adata_out.labels = net_adata.obs['labels']
+
```
romain-lopez, line 211: Usually, we learn scVI first, and then on the initialized weights of scVI run the scANVI training. I need Pierre to verify this.

scIB/integration.py
```diff
@@ -174,6 +174,69 @@ def runScvi(adata, batch, hvg=None):
...
+        raise TypeError('Adata does not contain a `counts` layer in `adata.layers[`counts`]`')
+
+    from scvi.models import VAE, SCANVI
+    from scvi.inference import AlternateSemiSupervisedTrainer, SemiSupervisedTrainer
+    from sklearn.preprocessing import LabelEncoder
+    from scvi.dataset import AnnDatasetFromAnnData
+
+    # Defaults from SCVI github tutorials scanpy_pbmc3k and harmonization
+    n_epochs=np.min([round((20000/adata.n_obs)*400), 400])
+    n_latent=30
```
romain-lopez, line 194: n_latent are defined here but not fed into the model. I guess it depends what you're evaluating.

scIB/integration.py
```diff
@@ -174,6 +174,69 @@ def runScvi(adata, batch, hvg=None):
...
+    net_adata_out = AnnDatasetFromAnnData(net_adata)
+    net_adata_out.labels = net_adata.obs['labels']
+
+    scanvi = SCANVI(
+        net_adata_out.nb_genes,
+        reconstruction_loss='nb',
+        n_batch=net_adata_out.n_batches,
+        n_labels=net_adata_out.n_labels,
+        n_layers=n_layers,
+        #n_latent=n_latent,
```
romain-lopez, line 218: those commented lines seem dangerous, see my other comment

scIB/integration.py
```diff
@@ -174,6 +174,69 @@ def runScvi(adata, batch, hvg=None):
...
+        #n_hidden=n_hidden,
+    )
+
+    trainer = SemiSupervisedTrainer(
+        scanvi,
+        net_adata_out,
+        train_size=1.0,
+        n_epochs_classifier=10,
```
romain-lopez, line 226: there are many trainers and parameters. I need Pierre to confirm which one to use (I am not sure myself)

scIB/integration.py
```diff
@@ -174,6 +174,69 @@ def runScvi(adata, batch, hvg=None):
...
+        use_cuda=False,
+    )
+    
+    trainer.labelled_set = trainer.create_posterior()
```
romain-lopez, line 230: That seems definitely super suspect. In scANVI you need to redefine what is the labelled set and what is the unlabelled set systematically. I don't know what this line would do exactly. Does it just use all cells as labelled? 

Which cell type information do you want to give to scANVI? All the cell labels? Or only a few of them? Do you have some "unlabelled" inside of the labels field?

Usually we use only label from one dataset for example or only a couple of labels from a single dataset. It does not seem meaningful to use scANVI like this? I'm unsure. Let me know if what I say makes sense.

scIB/integration.py
```diff
@@ -174,6 +174,69 @@ def runScvi(adata, batch, hvg=None):
...
+
+    trainer.train(n_epochs=10)
```
romain-lopez, line 232: same here, n_epochs is defined but is not used. Also, I'll need Pierre here to confirm how to adjust this number.

scIB/integration.py
danielStrobl: Hey! Thanks for your comments!

We would want to all of the labels that we have for integration, we would not have any unlabelled cells. What we basically wanted to to is to use scANVI in a similar way to scVI but with added cell type labels.
I'm not really sure if we can use scANVI in such a way.

scIB/integration.py
LuckyMD: I see this in the tutorial now, yes... this should be added. Thanks for pointing it out. I guess that makes both methods intertwined (as they are anyway), but that's fine if this is the canonical way to run scANVI.

scIB/integration.py
LuckyMD: The though process here is that adding more prior information, on which you condition your latent variable z will improve the latent representation. So we pass all labels to get the best possible representation. If this approach doesn't make sense to you, maybe we can also omit scANVI and only compare scVI though, as then conceptually there might not be much difference except the added layer of complexity.

romain-lopez: Hi Malte, 

I am going to add in this PR a piece of code to do that. The idea is that the labels should help structure the latent space better, and be some form of additional information. 

One thing that would be good to add is that usually, scANVI is used for probabilistic annotations of cells, and when it is fed the complete annotation, it won't be probabilistic in the cell types anymore. Consequently, the latent space becomes biased with respect to the fed labels. Hopefully, this model should get better clustering results than scVI for example. It might deform the latent space in other ways also. 

Also, I would insist on the fact that usually not all the labels are given. But I agree this is nice to add it as a "cell type informed" latent space and see what comes out of it.

scIB/integration.py
romain-lopez: yes, we have been experimenting with training procedures for quite a while. This is the most stable way of running the method.

romain-lopez:
Hi, please find a code snippet below, on how to run scanvi properly in this setting
```

def runScanvi(adata, batch, labels, hvg=True):
    # Use non-normalized (count) data for scanvi!
    
    # Check for counts data layer
    if 'counts' not in adata.layers:
        raise TypeError('Adata does not contain a `counts` layer in `adata.layers[`counts`]`')

    from scvi.models import VAE, SCANVI
    from scvi.inference import UnsupervisedTrainer, SemiSupervisedTrainer
    from sklearn.preprocessing import LabelEncoder
    from scvi.dataset import AnnDatasetFromAnnData
    
    # STEP 1: prepare the data
    net_adata = adata.copy()
    net_adata.X = adata.layers['counts']
    del net_adata.layers['counts']
    # Ensure that the raw counts are not accidentally used
    del net_adata.raw # Note that this only works from anndata 0.7

    # Define batch indices
    le = LabelEncoder()
    net_adata.obs['batch_indices'] = le.fit_transform(net_adata.obs[batch].values)
    net_adata.obs['labels'] = le.fit_transform(net_adata.obs[labels].values)

    net_adata = AnnDatasetFromAnnData(net_adata)

    print("scANVI dataset object with {} batches and {} cell types".format(net_adata.n_batches, net_adata.n_labels))

    if hvg is True:
        # this also corrects for different batches by default
        net_adata.subsample_genes(2000, mode="seurat_v3")

    # # Defaults from SCVI github tutorials scanpy_pbmc3k and harmonization
    n_epochs_scVI = np.min([round((20000/adata.n_obs)*400), 400])
    n_epochs_scANVI = np.min([10, np.max([2, round(n_epochs_scVI / 3.)])])
    n_latent=30
    n_hidden=128
    n_layers=2


    # STEP 2: RUN scVI to initialize scANVI

    vae = VAE(
        net_adata.nb_genes,
        reconstruction_loss='nb',
        n_batch=net_adata.n_batches,
        n_latent=n_latent,
        n_hidden=n_hidden,
        n_layers=n_layers,
    )

    trainer = UnsupervisedTrainer(
        vae,
        net_adata,
        train_size=1.0,
        use_cuda=True,
    )

    trainer.train(n_epochs=n_epochs_scVI, lr=1e-3)

    # STEP 3: RUN scANVI

    scanvi = SCANVI(net_adata.nb_genes, net_adata.n_batches, net_adata.n_labels,
                          n_hidden=n_hidden, n_latent=n_latent, n_layers=n_layers, dispersion='gene')
    scanvi.load_state_dict(trainer.model.state_dict(), strict=False)

    # use default parameter from semi-supervised trainer class
    trainer_scanvi = SemiSupervisedTrainer(scanvi, net_adata)
    # use all cells as labelled set
    trainer_scanvi.labelled_set = trainer_scanvi.create_posterior(trainer_scanvi.model, net_adata, indices=np.arange(len(net_adata)))
    # put one cell in the unlabelled set
    trainer_scanvi.unlabelled_set = trainer_scanvi.create_posterior(indices=[0])
    trainer_scanvi.train(n_epochs=n_epochs_scANVI)

    # extract info from posterior
    scanvi_full = trainer_scanvi.create_posterior(trainer_scanvi.model, net_adata, indices=np.arange(len(net_adata)))
    latent, _, _ = scanvi_full.sequential().get_latent()

    adata.obsm['X_emb'] = latent

    return adata, scanvi_full.sequential(), trainer_scanvi
```

romain-lopez:
Ah, I would check this carefully, I have changed a few things while testing inadvertantly like "hvg=True"

danielStrobl:
tested and it's working fine

scIB/integration.py
```diff
@@ -192,7 +192,88 @@ def runScvi(adata, batch, hvg=None):
     adata.obsm['X_emb'] = latent
 
     return adata
+def runScanvi(adata, batch, labels, hvg=None):
```
LuckyMD, line 195: `hvg` is not used in the function. Could remove this argument.

scIB/integration.py
```diff
@@ -192,7 +192,88 @@ def runScvi(adata, batch, hvg=None):
...
+
+    print("scANVI dataset object with {} batches and {} cell types".format(net_adata.n_batches, net_adata.n_labels))
+
+    #if hvg is True:
+    #    # this also corrects for different batches by default
+    #    net_adata.subsample_genes(2000, mode="seurat_v3")
+
+    # # Defaults from SCVI github tutorials scanpy_pbmc3k and harmonization
+    n_epochs_scVI = np.min([round((20000/adata.n_obs)*400), 400]) #400
+    n_epochs_scANVI = np.min([10, np.max([2, round(n_epochs_scVI / 3.)])])
```
LuckyMD, line 230: `n_epochs_scANVI` seems very low as a default. This is definitely okay @romain-lopez?

scIB/integration.py
```diff
@@ -192,7 +192,88 @@ def runScvi(adata, batch, hvg=None):
...
+        n_latent=n_latent,
+        n_hidden=n_hidden,
+        n_layers=n_layers,
+    )
+
+    trainer = UnsupervisedTrainer(
+        vae,
+        net_adata,
+        train_size=1.0,
+        use_cuda=True,
```
LuckyMD, line 251: `use_cuda` sounds like GPU, which we don't have in our setup on the server. Does it still work without GPU available? We have `use_cude=False` for scVI defaults in the `runScvI()` function.

scIB/integration.py
```diff
@@ -192,7 +192,88 @@ def runScvi(adata, batch, hvg=None):
...
+    trainer_scanvi.train(n_epochs=n_epochs_scANVI)
+
+    # extract info from posterior
+    scanvi_full = trainer_scanvi.create_posterior(trainer_scanvi.model, net_adata, indices=np.arange(len(net_adata)))
+    latent, _, _ = scanvi_full.sequential().get_latent()
+
+    adata.obsm['X_emb'] = latent
 
+    return adata#, scanvi_full.sequential(), trainer_scanvi
 def runMNN(adata, batch, hvg = None):
```
LuckyMD, line 277: add 2 new lines here

LuckyMD (changes requested):
In general it looks good, but some changes needed: don't scale input data, `use_cuda` default set to True, and a bit of cleanup.

LuckyMD:
> tested and it's working fine

tested in Snakemake or separately?

scIB/integration.py
danielStrobl: that's true for almost every function I think

scIB/integration.py
danielStrobl: That's the value he proposed here: https://github.com/theislab/scib/pull/132#issuecomment-668870534

scIB/integration.py
danielStrobl: I've run it on `icb-sarah`which doesn't have a GPU as far as I know, but yes, setting this to False might be sensible

config.yaml
```diff
@@ -58,6 +58,9 @@ METHODS: # method name : ouput type(s)
   scgen:
     output_type: full
     use_celltype: true
+  scanvi:
+    output_type: embed
+    use_celltype: true
```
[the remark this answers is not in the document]
danielStrobl: yes, we should deactivate scaling here I think

danielStrobl:
> > tested and it's working fine
> 
> tested in Snakemake or separately?

tested the `runIntegration.py`script

scIB/integration.py
LuckyMD: okay, let's go with it!

scIB/integration.py
LuckyMD: legacy argument i guess... let's remove here though... as no reason to add to the legacy code.

scIB/integration.py
LuckyMD: maybe test briefly via snakemake cpu queue as well.

danielStrobl:
ok, will quickly test on the queue and then it should be ready to merge I guess

LuckyMD (approved):
looks great now! Confirm queue test results are good and let's merge!

scIB/integration.py
LuckyMD: I will be careful to mention this in the methods section and if space permits also allude to it in the discussion. Thanks for your help, @romain-lopez!

romain-lopez: Great thanks!

danielStrobl:
Tested on snakemake and metrics test run completed now

PR #133, opened 2020-08-03
Unintegrated metrics

mumichae:
Fixed snakemake structure to allow for running metrics on unintegrated data.

- The unintegrated adata was preprocessed with `runPP.py` prior to metric computation. This was not the case before.
- unintegrated metrics are done for each scaling, feature subset and scenario
- unintegrated metrics files separated from integrated metrics files  `{scenario}/metrics_unintegrated/`
- unintegrated metrics are added to the final `metrics.csv` file

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -4,6 +4,8 @@ py_env : scIB-python
 conv_env : scIB-R-conversion
 timing: false
 
+unintegrated_metrics: false
+
 FEATURE_SELECTION:
   hvg: 2000
   full_feature: 0
--- a/scIB/preprocessing.py
+++ b/scIB/preprocessing.py
@@ -339,11 +339,11 @@ def hvg_batch(adata, batch_key=None, target_genes=2000, flavor='cell_ranger', n_
 ### Feature Reduction
 def reduce_data(adata, batch_key=None, subset=False,
                 filter=True, flavor='cell_ranger', n_top_genes=2000, n_bins=20,
-                pca=True, pca_comps=50, ignore_hvg=True,
+                pca=True, pca_comps=50, overwrite_hvg=True,
                 neighbors=True, use_rep='X_pca', 
                 umap=True):
     """
-    ignore_hvg:
+    overwrite_hvg:
         if True, ignores any pre-existing 'highly_variable' column in adata.var
         and recomputes it if `n_top_genes` is specified else calls PCA on full features.
         if False, skips HVG computation even if `n_top_genes` is specified and uses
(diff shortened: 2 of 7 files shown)


LuckyMD:
This was still WIP, right @mumichae ?

scripts/metrics.py
```diff
@@ -136,6 +139,11 @@
         # legacy check
         if ('emb' in adata_int.uns) and (adata_int.uns['emb']):
             adata_int.obsm["X_emb"] = adata_int.obsm["X_pca"].copy()
+        # fix for unintegrated metrics
+        if "X_emb" not in adata_int.obsm:
```
LuckyMD, line 143: why is this necessary? Can't you just encode unintegrated not as an embedding methods, but as `type=full`?

scripts/metrics.py
```diff
@@ -244,11 +252,11 @@
                               lisi_graph_= lisi_graph_,
                               trajectory_=trajectory_
                              )
-    results.rename(columns={results.columns[0]:out_prefix}, inplace=True)
+    results.rename(columns={results.columns[0]:setup}, inplace=True)
     if verbose:
         print(results)
     # save metrics' results
-    results.to_csv(os.path.join(args.output, f'{out_prefix}.csv'))
+    results.to_csv(args.output)
```
LuckyMD, line 259: why do you no longer need the filename here?

scIB/preprocessing.py
```diff
@@ -339,17 +339,26 @@ def hvg_batch(adata, batch_key=None, target_genes=2000, flavor='cell_ranger', n_
...
     checkAdata(adata)
     if batch_key:
         checkBatch(batch_key, adata.obs)
     
-    hvg = n_top_genes is not None
-    if hvg:
+    if n_top_genes is not None and ignore_hvg:
         print("HVG")
+        
+        ignore_hvg = False
```
LuckyMD, line 360: why is this needed?

config.yaml
```diff
@@ -4,6 +4,8 @@ py_env : scIB-python
 conv_env : scIB-R-conversion
 timing: false
 
+unintegrated_metrics: false
```
LuckyMD, line 7: this is to toggle whether to run unintegrated metrics?

scripts/metrics.py
mumichae: script now expects filename as input instead of directory. Should be the same as for any other rule

config.yaml
mumichae: yes

mumichae: if not specified, will be interpreted as `False`

PR #137, opened 2020-08-04
Rescale lisi

mbuttner:
Rescaling of LISI score with nbatches and disabling the scale LISI rule in snakemake

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -205,16 +205,16 @@ rule cc_single:
         --hvgs {params.hvgs} --organism {params.organism}
         """
 
-rule scale_lisi:
-    input:
-        i = cfg.get_filename_pattern("metrics", "final"),
-        script = "scripts/scale_halfopen.py"
-    output: cfg.get_filename_pattern("metrics", "scaled_final")
-    message: "Rescale LISI in merged metrics"
-    params:
-        cmd = f"conda run -n {cfg.py_env} python"
-    shell:
-        """
-        {params.cmd} {input.script} -i {input.i} -o {output}
-        """
+#rule scale_lisi:
+#    input:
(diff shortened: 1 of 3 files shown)


Snakefile
```diff
@@ -205,16 +205,16 @@ rule cc_single:
         --hvgs {params.hvgs} --organism {params.organism}
         """
 
-rule scale_lisi:
```
LuckyMD, line 208: If you remove this rule, you will also need to change `all:` to accept the output from the `metrics:` rule as output.

mbuttner:
> Snakemake consistency doesn't work without modification of `all` i think. Also it needs testing I assume?

I'm not familiar enough with snakemake to follow your requests, unfortunately.

mbuttner:
> Also it needs testing I assume?

I tested the rescaling code in the lisi function in metrics.py (with 1 test file (pancreas - LIGER) in a jupyter notebook - lisi scores behave as expected). I do not know how to test the changes to the snakemake, though.

LuckyMD:
Would need to use the `snakemake_test` environment and create a suitable `config.yaml` file to only test 1 or 2 cases to see that it works.

Snakefile
mbuttner: Change `all:` rule in commit (a37f239 )[https://github.com/theislab/scib/pull/137/commits/a37f239ca98a282f66a660bc20fabb9c4e2199e9]

mumichae: The change in all is the essential change for snakemake :+1: . The rest is just cleaning up.

mbuttner:
I rescaled the lisi graph function now in commit [8df0bd2](https://github.com/theislab/scib/pull/137/commits/8df0bd21ad3a3946462eb25d07df771daef8e2ea)

PR #139, opened 2020-08-05
Error for saucie if input is dense

danielStrobl:
The saucie function didn't work with dense input, this fixes that


scIB/integration.py
```diff
@@ -220,7 +220,10 @@ def runSaucie(adata, batch):
     import SAUCIE
     import sklearn.decomposition
     pca_op = sklearn.decomposition.PCA(100)
-    expr = adata.X.todense()
+    if isinstance(adata.X, sp.sparse.csr_matrix):
+        expr = adata.X.todense()
```
LuckyMD, line 224: this should probably be a `adata.X.toarray()` or a `adata.X.A`, no?

PR #144, opened 2020-08-23
scIB not compatible with scanpy >= 1.5

scottgigante:


LuckyMD:
Thanks, @scottgigante !

LuckyMD (approved):

PR #147, opened 2020-08-27
Update scIB-python.yml

danielStrobl:
Updated the python env


envs/scIB-python.yml
```diff
@@ -5,508 +5,58 @@ channels:
...
+  - numpy
+  - pandas
+  - seaborn
+  - matplotlib
+  - scanpy==1.4.6
+  - scipy
+  - memory_profiler
+  - rpy2>=3
+  - anndata2ri
+  - scanorama
```
LuckyMD, line 17: would be more comfortable with a version here to show we control which version we are using to benchmark

envs/scIB-python.yml
```diff
@@ -5,508 +5,58 @@ channels:
...
+  - bbknn==1.3.9
+  - libgcc-ng
+  - scikit-learn
+  - networkx
+  - r-base
+  - r-seurat
```
LuckyMD, line 23: also add version, please... or maybe it matters less here, as it's not the `R` env.

envs/scIB-python.yml
```diff
@@ -5,508 +5,58 @@ channels:
...
-    - torch==1.2.0
-    - torchvision==0.5.0
-    - trvae==0.0.1
-    - trvaep==0.1.0
-    - urllib3==1.25.3
-    - xlrd==1.2.0
-    - zmq==0.0.0
-prefix: /home/icb/daniel.strobl/miniconda3/envs/sc-tutorial
-
+    - trvae
```
LuckyMD, line 53: add version, please :)

envs/scIB-python.yml
```diff
@@ -5,508 +5,58 @@ channels:
...
+    - trvaep
+    - mnnpy
```
LuckyMD, line 55: again, version

LuckyMD (changes requested):
just need to add version numbers to some tools

LuckyMD:
@danielStrobl There doesn't seem to be any `kBet` in here...

envs/scIB-python.yml
```diff
@@ -48,11 +47,13 @@ dependencies:
...
   - xlrd
+  - umap-learn==0.3.10
   - pip:
     - anndata==0.7.1
-    - trvae
-    - trvaep
-    - mnnpy
+    - trvae==1.1.2
+    - trvaep==0.1.0
+    - scanorama+git://github.com/danielStrobl/scanorama@patch-1
```
lazappi, line 55: I had to change this to `git+git://github.com/danielStrobl/scanorama@patch-1` to get it to work.

lazappi:
There are also a couple of things that are linux specific and need to be removed to install on MacOS

danielStrobl:
This should be merge ready now. Tested on my local machine. Also, manually setting the .libPaths() is not necessary anymore due to the inclusion of `r-environment` which seems to do this automatically

README.md
```diff
@@ -36,6 +36,9 @@ conda env create -f FILENAME.yml
 ``` 
 To set the correct paths so that R the correct R libraries can be found, copy `env_vars_activate.sh` to `etc/conda/activate.d/`
 and `env_vars_deactivate.sh` to `etc/conda/deactivate.d/` to every environment.
+As `Saucie` is not available as a proper python package for the scIB-python env, it needs to be installed manually from [here](https://github.com/KrishnaswamyLab/SAUCIE).
```
LuckyMD, line 39: remove the word "proper".

envs/scIB-python.yml
```diff
@@ -5,508 +5,63 @@ channels:
...
+  - scipy
+  - memory_profiler
+  - rpy2>=3
+  - anndata2ri
+  - bbknn==1.3.9
+  - libgcc-ng
+  - scikit-learn
+  - networkx
+  - r-base
+  - r-seurat==3.0.2
```
LuckyMD, line 22: not 3.1.1? We use this version for integration in the paper.

envs/scIB-python.yml
```diff
@@ -5,508 +5,63 @@ channels:
...
+  - r-fnn
+  - r-isoband
+  - r-lifecycle
+  - r-praise
+  - r-testthat
+  - r-vctrs
+  - xlrd
+  - umap-learn==0.3.10
+  - louvain==0.6.1
+  - scvi
```
LuckyMD, line 53: this is the only method that is not version-controlled. Should this be pinned too? I think it was version 0.6.6

envs/scIB-python.yml
```diff
@@ -5,508 +5,63 @@ channels:
...
-    - scipy==1.2.1
-    - scvi==0.5.0
-    - sklearn==0.0
-    - sortedcontainers==2.1.0
-    - termcolor==1.1.0
-    - torch==1.2.0
-    - torchvision==0.5.0
-    - trvae==0.0.1
+    - trvae==1.1.2
     - trvaep==0.1.0
```
LuckyMD, line 57: can this be installed? Also, trvae 1.1.2 is installable, but we can't use it without the latest PR.

envs/scIB-python.yml
```diff
@@ -5,508 +5,63 @@ channels:
...
-    - urllib3==1.25.3
-    - xlrd==1.2.0
-    - zmq==0.0.0
-prefix: /home/icb/daniel.strobl/miniconda3/envs/sc-tutorial
-
+    - scgen==1.1.5
+    - git+git://github.com/danielStrobl/scanorama@patch-1
```
LuckyMD, line 59: is this still needed? I thought it's been patched now?

LuckyMD (changes requested):
A few package version questions still. @lazappi and @mumichae you guys both tried to install this env. Anything missing here from your sides?

LuckyMD:
> Also, manually setting the .libPaths() is not necessary anymore due to the inclusion of `r-environment` which seems to do this automatically

That's very cool!

envs/scIB-python.yml
danielStrobl: I don't think that really matters here as it's just for conversion. I can try with the 3.1.1 though

envs/scIB-python.yml
danielStrobl: Yes, this sould be 0.6.7

envs/scIB-python.yml
danielStrobl: Yes, but I haven't tested with the new version yet. This is the one that we used for integration so I thought it would be ok to leave it like that

envs/scIB-python.yml
LuckyMD: yes, let's be consistent.

envs/scIB-python.yml
LuckyMD: okay, please adapt :)

envs/scIB-python.yml
LuckyMD: hmm... let's update and check for consistency. Always better to rely on the official version

lazappi:
> A few package version questions still. @lazappi and @mumichae you guys both tried to install this env. Anything missing here from your sides?

I've forgotten all the issues I ran into. I think I gave @danielStrobl a list at some point, if those have been addressed probably good from my side.

envs/scIB-python.yml
mumichae: The newest version is 1.7, Daniel's patch is on version 1.4; which version should we include for this environment?

LuckyMD: can we check whether there are consistent results between the two versions? If yes, then we can go with 1.7

envs/scIB-python.yml
```diff
@@ -5,508 +5,63 @@ channels:
...
+  - scikit-learn
+  - networkx
+  - r-base
+  - r-seurat==3.0.2
+  - bioconductor-scater
+  - bioconductor-scran
+  - pip
+  - numba<=0.46
+  - llvmlite
+  - tensorflow==1.15
```
LuckyMD, line 28: are we sure this is okay? The previous env had 1.1.4 and this version clashes between scGen and DESC I believe.

envs/scIB-python.yml
```diff
@@ -5,508 +5,63 @@ channels:
...
-
+    - scgen==1.1.5
+    - git+git://github.com/danielStrobl/scanorama@patch-1
+    - mnnpy==0.1.9.5
+    - git+git://github.com/theislab/scib.git  
+    - desc==2.0.3
+    - pydot
+    - loompy==3.0.6
+    - scikit-misc
+    - keras==2.1
```
LuckyMD, line 66: also this was 2.2.4 in the previous env (and clash for DESC and scGen). Why change now?

LuckyMD (changes requested):

danielStrobl:
I tested to install the env with the recent changes on my local machine, the installation ran through without errors and `pip check` also doesn't return errors after installation.

scIB/integration.py
```diff
@@ -28,10 +28,12 @@ def runScanorama(adata, batch, hvg = None):
...
     checkSanity(adata, batch, hvg)
     split = splitBatches(adata.copy(), batch)
-    emb, corrected = scanorama.correct_scanpy(split, return_dimred=True)
+    corrected = scanorama.correct_scanpy(split, return_dimred=True)
     corrected = corrected[0].concatenate(corrected[1:])
-    emb = np.concatenate(emb, axis=0)
-    corrected.obsm['X_emb']= emb
+    print(corrected.obs)
+    print(corrected.obsm)
+    print(corrected.var)
```
LuckyMD, line 35: could we remove the print statements here? @mumichae

envs/scIB-R-integration.yml
```diff
@@ -284,7 +287,7 @@ dependencies:
     - pytest==5.3.1
     - pytz==2019.3
     - rpy2==3.0.5
-    - scanpy==1.4.5.dev114+gd69832a
+    - scanpy==1.4.5
```
LuckyMD, line 290: does this downgrade affect anything important in our pipeline? Maybe better to use a git install, no?

envs/scIB-python.yml
```diff
@@ -50,18 +53,16 @@ dependencies:
...
     - trvae==1.1.2
     - trvaep==0.1.0
-    - scgen==1.1.5
-    - git+git://github.com/danielStrobl/scanorama@patch-1
     - mnnpy==0.1.9.5
-    - git+git://github.com/theislab/scib.git  
-    - desc==2.0.3
+      # - scgen==1.1.5
+      # - desc==2.0.3
+      # - keras==2.2.4
```
LuckyMD, line 65: could we please keep scgen in the environment? It's desc and scgen that clash, so we can use one over the other. As desc doesn't perform well, let's opt for scgen.

LuckyMD (changes requested):
a few small changes, then it's good to go @mumichae

envs/scIB-R-integration.yml
mumichae: not sure where this version comes from and if there was any  new feature that we needed @danielStrobl Do you remember how we installed it and why we didn't use the pip version?
I'm guessing that we might have had some computational issues, but `scanpy==1.4.5` has been working well so far. It's also the same version that we're using in the python environment

mumichae: If anything, I think we should have same versions for the python and R environments, i.e. 1.4.6

danielStrobl: I think there was one bug in the Scanpy code that we wanted to avoid, so we fixed it to this version. Unfortunately the slack messages for this are long gone.

envs/scIB-python.yml
danielStrobl: This looks like this was mistakenly commented out, but not sure. I think you can install desc first and then scgen after which would install the dependencies for scgen while ignoring the desc ones

LuckyMD: Okay, could you suggest the fix and then both test locally? I think @mumichae had an issue with this, no?

envs/scIB-R-integration.yml
LuckyMD: 1.4.6 sounds good to me. Could you adapt and test @mumichae ?

scIB/integration.py
mumichae: thanks, forgot about them

envs/scIB-python.yml
mumichae: scgen also clashes with trvae
```
The conflict is caused by:
    scgen 1.1.5 depends on keras==2.3
    trvae 1.1.2 depends on keras==2.2.4
```

LuckyMD: scgen > desc > trvae in importance here atm i think.

mumichae: added scgen and removed trvae

envs/scIB-R-integration.yml
mumichae: Updating to 1.4.6 didn't change the metrics. Most likely scanpy isn't used in this environment at all.

LuckyMD (approved):

LuckyMD:
Well done @mumichae and @danielStrobl! Great to finally have this merged!

PR #149, opened 2020-08-27
added log-transform argument to normalize function

LisaSikkema:
good to make explicit that log-transformation is performed. Set default to True, for compatibility with rest of scripts.

### DIFF
--- a/scIB/preprocessing.py
+++ b/scIB/preprocessing.py
@@ -143,7 +143,7 @@ def plot_count_filter(adata, obs_col='n_counts', bins=60, lower=0, upper=np.inf,
         plt.show()
 
 ### Normalisation
-def normalize(adata, min_mean = 0.1):
+def normalize(adata, min_mean = 0.1, logtransform=True):
     
     checkAdata(adata)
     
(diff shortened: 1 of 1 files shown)


danielStrobl (approved):
Looks good!

scIB/preprocessing.py
```diff
@@ -173,7 +173,11 @@ def normalize(adata, min_mean = 0.1):
     # modify adata
     adata.obs['size_factors'] = size_factors
     adata.X /= adata.obs['size_factors'].values[:,None]
-    sc.pp.log1p(adata)
+    if logtransform == True:
```
danielStrobl, line 176: I think you could just do `if logtransform:` here

scIB/preprocessing.py
```diff
@@ -143,7 +143,7 @@ def plot_count_filter(adata, obs_col='n_counts', bins=60, lower=0, upper=np.inf,
         plt.show()
 
 ### Normalisation
-def normalize(adata, min_mean = 0.1):
+def normalize(adata, min_mean = 0.1, logtransform=True):
```
LuckyMD, line 146: could we turn this argument into the simpler `log=True`?

LuckyMD (changes requested):
a little simplification of the code needed :)

LuckyMD (approved):

LuckyMD:
Thanks a lot @LisaSikkema !

PR #154, opened 2020-09-07
Fixing F1 isolated label score

mumichae:
added isolated label fix (argmax indexing)


LuckyMD (approved):

LuckyMD:
This looks like a bug only for a case that we weren't using though, right? `argmax=False` in the defaults made sure of that...

Issue #155, opened 2020-09-10
Memory leak in scran normalization


scottgigante:
Seems that scran normalization is a) densifying the data and b) not freeing it afterwards.

```
import openproblems
import psutil
import gc
import time
import scIB


def mem_available():
    time.sleep(1)
    gc.collect()
    time.sleep(1)
    return psutil.virtual_memory().available


initial_memory = mem_available()
adata = openproblems.data.zebrafish.load_zebrafish(test=True)
post_load_memory = mem_available()
del adata
post_delete_memory = mem_available()

adata = openproblems.data.zebrafish.load_zebrafish(test=True)
pre_norm_memory = mem_available()
scIB.preprocessing.normalize(adata)
post_norm_memory = mem_available()
del adata
post_norm_delete_memory = mem_available()

import matplotlib.pyplot as plt
import scprep
fig, ax = plt.subplots()
ax.plot(
    range(6), 
    (
        initial_memory,
        post_load_memory,
        post_delete_memory,
        pre_norm_memory,
        post_norm_memory,
        post_norm_delete_memory,
    )
)

scprep.plot.tools.label_axis(
    ax.xaxis, 
    ticks=range(6),
    ticklabels=[
        "initial",
        "load",
        "delete",
        "reload",
        "norm",
        "delete",
    ],
    ticklabel_rotation=45, 
    ticklabel_horizontal_alignment='right'
)

fig.tight_layout()
```

Plot shows available memory at each stage of the process.
![mem_usage](https://user-images.githubusercontent.com/8499679/92754239-a5a5ce80-f358-11ea-9b7e-be45b0347471.png)

Related: https://github.com/singlecellopenproblems/SingleCellOpenProblems/issues/54

LuckyMD:
Hmm, thanks for the bug report. This likely has to do with using an `R` function from python. As `scran` is run in `R`, a copy of the data matrix is made and converted into an `R` matrix object. That object is not deleted at the moment, and therefore the memory is never freed. The `R` kernel is also probably still running the background.

I'm not sure how to close the `R` kernel, but we should be able to delete the `R` object after moving it back at least.

scottgigante:
This seems like a workable solution:

```
def r_garbage_collection():
    rpy2.robjects.r('rm(list=ls())')
    rpy2.robjects.r('gc()')
```

![image](https://user-images.githubusercontent.com/8499679/92775654-96308080-f36c-11ea-97a3-f398a0bab247.png)

[...]

PR #156, opened 2020-09-10
[WIP] Free memory after running normalize

LuckyMD:
This PR relates to issue #155 

Running `scIB.pp.normalize()` starts an `R` kernel which requires memory that is not freed after running the function. 

Several things to try:
- [x] Deleting the objects moved to `R` 
- [x] Unloading modules in `R`
- [ ] Closing down the `R` kernel after the computation is done

### DIFF
--- a/scIB/preprocessing.py
+++ b/scIB/preprocessing.py
@@ -185,6 +185,8 @@ def normalize(adata, min_mean = 0.1, log=True):
 
     # Free memory in R
     ro.r('rm(list=ls())')
+    ro.r('invisible(lapply(paste0("package:", names(sessionInfo()$otherPkgs)), '
+         'detach, character.only=TRUE, unload=TRUE))')
     ro.r('gc()')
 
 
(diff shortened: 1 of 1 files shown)


LuckyMD:
So removing the object and calling the garbage collector in `R` doesn't change anything:

Screenshot 2020-09-10 at 19 51 44

LuckyMD:
@scottgigante could you check if it works any better with modules unloaded? My memory use is not making a lot of sense...

LuckyMD:
Without unloading modules:
Screenshot 2020-09-11 at 19 12 18

With unloading modules:
Screenshot 2020-09-11 at 19 12 25

Very confusing...

scottgigante:
Note that memory use is system-wide so if there's anything else going on that will affect results. I'll try again with the new code when I get a chance to close everything else down.

LuckyMD:
Yeah, of course... i have a lot of tabs open... but haven't really been doing anything in between. Maybe some system job.

LuckyMD:
this seems to have fixed things. Tested and ready to merge.

PR #161, opened 2020-09-11
Catch metrics errors

danielStrobl:
Catch metrics errors arising for low quality input.
Not tested yet

### DIFF
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -1671,6 +1671,10 @@ def kBET(adata, batch_key, label_key, embed='X_pca', type_ = None,
     
     return kBET_scores
 
+Class RootCellError(Exception):
+    def __init__(self, message):
+        self.message = message
+
 # determine root cell for trajectory conservation metric
 def get_root(adata_pre, adata_post, ct_key, dpt_dim=3):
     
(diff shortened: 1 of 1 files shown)


LuckyMD:
I think I would go deeper to catch these errors. Especially a `ValueError` is probably something that is quite generic and so should be caught closer to the source. That being said... we might just set a threshold for the connectivity of the subsetted knn graph.

danielStrobl:
> I think I would go deeper to catch these errors. Especially a `ValueError` is probably something that is quite generic and so should be caught closer to the source. That being said... we might just set a threshold for the connectivity of the subsetted knn graph.

I'm not sure if that's possible, this is the error message that is raised:
```
raise ValueError(f'could not find {k} nearest neighbors in {max_iterations}'
ValueError: could not find 70 nearest neighbors in 26diffusion steps.
 Please increase max_iterations or reduce k.
```

LuckyMD:
Super nice solution now. Very elegant!

LuckyMD (approved):
This looks super nice! Only maybe collect the new error class definition at the top or bottom of the file.

danielStrobl:
Done!

LuckyMD:
tested and it works?

PR #165, opened 2020-09-17
changed trajectory metric to only consider largest connected component

danielStrobl:
Only use the largest connected component for dpt calculation in post integration object.
Tested in a jupyter notebook so far.

### DIFF
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -1676,7 +1676,7 @@ def get_root(adata_pre, adata_post, ct_key, dpt_dim=3):
     max_neigh = adata_post.obs[adata_post.obs['neighborhood']== adata_post.obs['neighborhood'].value_counts().idxmax()].index
     min_dpt = [value for value in min_dpt if value in max_neigh]
     
-    adata_post_sub = adata_post[adata_post.obs['neighborhood']== adata_post.obs['neighborhood'].value_counts().argmax()]
+    adata_post_sub = adata_post[adata_post.obs['neighborhood']== adata_post.obs['neighborhood'].value_counts().idxmax()]
     
     min_dpt = [adata_post_sub.obs_names.get_loc(i) for i in min_dpt]
     
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -747,7 +747,7 @@ def select_hvg(adata, select=True):
         return adata
 
 ### diffusion for connectivites matrix extension
-def diffusion_conn(adata, min_k=50, copy=True, max_iterations=20):
+def diffusion_conn(adata, min_k=50, copy=True, max_iterations=26):
     '''
     This function performs graph diffusion on the connectivities matrix until a
     minimum number `min_k` of entries per row are non-zero.
@@ -815,7 +815,7 @@ def diffusion_conn(adata, min_k=50, copy=True, max_iterations=20):
 
     
 ### diffusion neighbourhood score
-def diffusion_nn(adata, k, max_iterations=20):
+def diffusion_nn(adata, k, max_iterations=26):
     '''
     This function generates a nearest neighbour list from a connectivities matrix
     as supplied by BBKNN or Conos. This allows us to select a consistent number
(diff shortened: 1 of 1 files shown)


danielStrobl:
Also tested the script now and it's working fine

LuckyMD:
Are results consistent with the previous version?

scIB/metrics.py
```diff
@@ -747,7 +747,7 @@ def select_hvg(adata, select=True):
         return adata
 
 ### diffusion for connectivites matrix extension
-def diffusion_conn(adata, min_k=50, copy=True, max_iterations=20):
+def diffusion_conn(adata, min_k=50, copy=True, max_iterations=26):
```
LuckyMD, line 750: do we want to change this default, or should we say if no result is reached, then we set the score to 0?

scIB/metrics.py
```diff
@@ -815,7 +815,7 @@ def diffusion_conn(adata, min_k=50, copy=True, max_iterations=20):
 
     
 ### diffusion neighbourhood score
-def diffusion_nn(adata, k, max_iterations=20):
+def diffusion_nn(adata, k, max_iterations=26):
```
LuckyMD, line 818: see above

scIB/metrics.py
```diff
@@ -1667,22 +1667,28 @@ def kBET(adata, batch_key, label_key, embed='X_pca', type_ = None,
...
 def get_root(adata_pre, adata_post, ct_key, dpt_dim=3):
+    
     n_components, adata_post.obs['neighborhood'] = connected_components(csgraph=adata_post.uns['neighbors']['connectivities'], directed=False, return_labels=True)
     
     start_clust = adata_pre.obs.groupby([ct_key]).mean()['dpt_pseudotime'].idxmin()
-    min_dpt = np.flatnonzero(adata_pre.obs[ct_key] == start_clust)
-    max_neigh = np.flatnonzero(adata_post.obs['neighborhood']== adata_post.obs['neighborhood'].value_counts().argmax())
+    min_dpt = adata_pre.obs[adata_pre.obs[ct_key] == start_clust].index
+    #print(min_dpt)
+    max_neigh = adata_post.obs[adata_post.obs['neighborhood']== adata_post.obs['neighborhood'].value_counts().idxmax()].index
```
LuckyMD, line 1676: this looks like it still does the same as above... why change?

scIB/metrics.py
```diff
@@ -1667,22 +1667,28 @@ def kBET(adata, batch_key, label_key, embed='X_pca', type_ = None,
...
     min_dpt = [value for value in min_dpt if value in max_neigh]
     
+    adata_post_sub = adata_post[adata_post.obs['neighborhood']== adata_post.obs['neighborhood'].value_counts().argmax()]
```
LuckyMD, line 1679: if argmax is deprecated and idxmax should be used instead, why not also here?

LuckyMD: could also save `.value_counts().idxmax()` somewhere, as you reuse it here.

danielStrobl:
> Are results consistent with the previous version?

In the cases that I've tested, yes

LuckyMD (changes requested):
1. A few changes to clean up the code. 
2. checks for consistency would be good as the results should be pretty much the same (before we had `Inf` values for dpt which we then set to give a 0 score contribution).
3. Decision on defaults for `max_iterations` or set score to 0.

scIB/metrics.py
danielStrobl: But how do we decide how to cut off? I guess fixing a value to cut off here would be pretty random. Doesn't seem to take super long either way

scIB/metrics.py
danielStrobl: the old one returned the integer index, the new one uses the obs_name. I think the integer index changes when we subset the post integration object to a different size as the pre_integration object so it would change. Another option would probably be to subset adata_pre to the largest connected component as well.

scIB/metrics.py
danielStrobl: ah, I missed that one

scIB/metrics.py
LuckyMD: Why did we ever work with the integer index? Good to fix this.

scIB/metrics.py
LuckyMD: doesn't take long? Then okay. Let's keep at 26 then... although the more iterations, the more we just smooth out the data and lose all signal.

LuckyMD (approved):

scIB/metrics.py
danielStrobl: I think this was due to changing `obs_names` between pre and post integration for some methods

scIB/metrics.py
danielStrobl: I realised this actually breaks the function if the largest component isn't 0

scIB/metrics.py
LuckyMD: I see. So as we have now fixed this in the metrics.py script, it doesn't matter anymore. Good stuff.

scIB/metrics.py
LuckyMD: isn't the largest component always 0? I think it should be ordered... but better to not make assumptions i guess.

PR #167, opened 2020-09-18
Modify how embeddings are stored in Seurat objects

### LINKED ISSUE #166: Error on R methods if there are all zero genes
If a dataset has all zero genes some of the R methods fail. The error comes from when an embedding is stored in a Seurat object, for example from `runHarm()`:

```r
sobj@reductions['X_emb'] <- sobj@reductions$harmony
```

This shouldn't be a problem if people filter their datasets properly but thought I should report it just in case it comes up. Possibly it can be fixed by using `sobj[['X_emb']]` instead but I haven't testing this.

lazappi:
For Harmony and fastMNN now store embeddings using `sobj[['X_emb']]` instead of `sobj@reductions['X_emb']`

Fixes #166 (hopefully)

### DIFF
--- a/R/integration.R
+++ b/R/integration.R
@@ -23,7 +23,7 @@ runSeurat = function(data, batch, hvg=2000) {
         	  k.score = 30,
         	  max.features = 200,
         	  eps = 0)
-	integrated = IntegrateData(
+	  integrated = IntegrateData(
         	   anchorset = anchors,
 		   new.assay.name = "integrated",
         	   features = NULL,
@@ -37,7 +37,7 @@ runSeurat = function(data, batch, hvg=2000) {
         	   do.cpp = T,
         	   eps = 0,
         	   verbose = T)
-	return(integrated)
+	  return(integrated)
 }
 func_profiler = function(expr, chunksize=20000, filename='timing.out', prof.interval=0.02) {
 	      Rprof(filename, memory.profiling=T, interval=prof.interval)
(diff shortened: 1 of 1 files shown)


danielStrobl (approved):
Looks good! Did you already test this in the snakemake pipeline?
If it's tested I think we could already merge this.

lazappi:
I've tested it using Snakemake on my laptop. Only on a simulated dataset and not ATAC. It's a pretty minor change though and seems to fix the error so shouldn't cause any issues. Not sure what @LuckyMD thinks?

R/integration.R
```diff
@@ -99,8 +99,8 @@ runHarm = function(sobj, batch) {
 	sobj <- ScaleData(sobj)
 	sobj <- RunPCA(sobj, features=rownames(sobj@assays$RNA))
 	sobj <- RunHarmony(sobj, batch)
-	sobj@reductions['X_emb'] <- sobj@reductions$harmony
-	#harmonyEmb <- HarmonyMatrix(pca, method, batch, do_pca=F)
+	sobj[['X_emb']] <- sobj[['harmony']]
+    #harmonyEmb <- HarmonyMatrix(pca, method, batch, do_pca=F)
```
LuckyMD, line 103: could we also remove the comment while we're at it?

LuckyMD (approved):
It looks good to me. Would be great to test this with snakemake, but i guess that'll happen anyway when we run the ATAC stuff on the server....

Just one clean-up please :)

R/integration.R
lazappi: This comment was already there. I'm happy to remove it but I think there are a few others in this file as well.

LuckyMD: yeah, i saw... community service: clean up after others :)

lazappi:
Cleaned up comments I didn't think were necessary. Thought about tidying up some other things but probably would have reformatted the whole file so decided to stop myself there 😸.

LuckyMD (approved):

PR #169, opened 2020-09-21
Add UMAP plots to metrics

lazappi:
Added a few lines to the `metrics.py` script to save UMAP plots coloured by batch and label. These are useful for being able to quickly see what the output from a method looks like. Might be more correct to make a new script/stage but this seemed like the quickest option.

I have tested this using the Snakemake pipeline with the methods I have working. Potentially there could be issues with other methods but hopefully not.


LuckyMD:
This would as standard produce .png images in the output directory. Maybe it would be better to do this in a `figures/` subdirectory that you create?

lazappi:
That would make sense I guess. Currently they are inside the `metrics` folder, so something like `metrics/scaled/hvg/method_type_batch.png`. Were you thinking add a new folder at this level like `metrics/scaled/hvg/figures/method_type_batch.png` or at the `metrics` level `figures/scaled/hvg/method_type_batch.png`?

I was also thinking it might be useful to just output the coordinates of each cell along with labels/batches. Not as good for a quick look but would make it easier to make custom plots. Any thoughts on that?

danielStrobl:
> I was also thinking it might be useful to just output the coordinates of each cell along with labels/batches. Not as good for a quick look but would make it easier to make custom plots. Any thoughts on that?
I think that's a really good idea, especially if we want to create vector graphics for the final figures.

Another point is that we can't use UMAPs for conos, we would need to use `sc.tl.draw_graph` for that case.

lazappi:
Yeah I thought conos might be a problem but I don't have it running to test. Do you have an example of what you used for that before?

danielStrobl:
I had this function for preprocessing in my visualisation notebook:
```
def pp(adatas, a_pre, batch_key):
    for key in adatas:
        print(f'preprocessing {key}...')
        ot = adatas[key][1]
        dat = adatas[key][0]
    
        # Ensure annotations are matched
        match_anns(dat, a_pre, batch_key)
    
        if ot == 'knn':
            if 'params' not in dat.uns['neighbors']:
                dat.uns['neighbors']['params'] = dict(n_pcs=dat.obsm['X_pca'].shape[1],
                                                      use_rep='X_pca',
                                                      metric='cosine',
                                                      method='umap',
                                                      n_neighbors=10)
            
            sc.tl.umap(dat)
        
        elif ot == 'emb':
            scIB.pp.reduce_data(dat, n_top_genes=None, neighbors=True, use_rep='X_emb', pca=True, umap=True)
        
        elif ot == 'gene':
            if key.startswith('tr'):
                sc.pp.filter_genes(dat, min_cells=1)
            scIB.pp.reduce_data(dat, n_top_genes=2000, neighbors=True, use_rep='X_pca', pca=True, umap=True)
        if key.startswith('conos'):
            print('conos comp graph')
            sc.tl.draw_graph(dat)
```

LuckyMD:
> Were you thinking add a new folder at this level like `metrics/scaled/hvg/figures/method_type_batch.png` or at the `metrics` level `figures/scaled/hvg/method_type_batch.png`?

I was thinking the former... the latter is harder to do and would require passing another argument to the metrics script via snakemake. The latter would be nicer though.

> I was also thinking it might be useful to just output the coordinates of each cell along with labels/batches.

that sounds good to me

Should we just add Daniel's function to the bottom of the metrics script? Or would you keep this separate?

danielStrobl:
I think the function would need to be adapted though as this was from a
jupyter notebook

lazappi:
I think that should be pretty close to what we want now. Still haven't been able to test it with CONOS but that's one of the next things on my list.

LuckyMD (approved):
all good if tested and it works.

lazappi:
Might be good if @danielStrobl or someone what has CONOS installed checks that works. Shouldn't be a problem but don't what to merge something I haven't tested.

LuckyMD:
agreed.

danielStrobl:
ok, I'll test that

lazappi:
@mumichae If you can suggest how to turn this into a Snakemake stage that would be great. I guess we would need to turn what I have added into a separate script and then add it to Snakemake somehow? I can probably do the script but need more help with Snakemake.

mumichae:
A script will be most of the work. The snakemake rule is pretty easy once we know  the input. Let me know when you are done with the script and I can integrate it

lazappi:
Ok. I think this is better now. Have tested it locally but still needs to be tested on CONOS output (@danielStrobl).

Here is what I have done:

* Moved the embedding plots out of `metrics.py` to a new `save_embeddings.py` script
* Reverted `metrics.py` to the existing version
* Added new stages to Snakemake that call the new script
* Output is now at `/dataset/embeddings/...` instead of in the `metrics` directory
* There is also a top level `embeddings.csv` file which just has the paths to the individual embedding CSV files. I couldn't work out how to get the `all` rule to call embedding without a summarised output but there is probably some way to avoid this. (Also possibly you don't want these to be created by all?)

@mumichae Can you please check the Snakemake stuff?

LuckyMD:
Do we want the `all` rule to create the embedded images? Or should this be a separate rule that you can call for post-processing?

Snakefile
```diff
@@ -207,14 +208,68 @@ rule metrics_single:
...
+        "Completed all embeddings"
+    shell:
+     """
+     echo '{input.csvs}' | tr " " "\n" > {output}
+     """
+
+rule embeddings_single:
+    input:
+        adata  = get_integrated_for_metrics,
+        script = "scripts/save_embeddings.py"
```
LuckyMD, line 252: why is this an input? Isn't this always the same script? Could then move this to just be hardcoded in params or in shell directly.

scripts/save_embeddings.py
```diff
@@ -0,0 +1,87 @@
...
+                            use_rep='X_pca', pca=True, umap=False)
+
+    # Calculate embedding
+    if args.method.startswith('conos'):
+        print('Calculating graph embedding...')
+        sc.tl.draw_graph(adata)
+        # May be named differently depending on packages so get the last key
+        graph_key = list(adata.obsm.keys())[-1]
+        basis = graph_key.lstrip('X_')
+        label = basis.replace('draw_graph_', '').upper()
```
LuckyMD, line 59: Not quite sure I understand this bottom part. Why not just make it clear what the draw_graph_output key should be by stating the `key_added_ext` parameter or sth like that. You can then search for the extension you chose. Are we sure that the order of `adata.obsm.keys()` always conforms to the last key being the most recently introduced key?

scripts/snakemake_parse.py
```diff
@@ -134,6 +135,7 @@ def get_filename_pattern(self, file_type, level, file_suffix=None):
             "prepare"     : "{method}.h5ad",
             "integration" : "{method}.h5ad",
             "metrics"     : "{method}_{o_type}.csv",
+            "embeddings"  : "{method}_{o_type}.csv",
             "cc_variance" : "{method}_{o_type}.csv"
```
LuckyMD, line 139: a little nerve-wrecking to have the same filename pattern 3 times here. I guess they are however stored in different folders, right? Is that also true for the `"cc_variance"` files here, @mumichae? Would be good to know if we could accidentally overwrite stuff otherwise.

LuckyMD (changes requested):
Looks very good overall. 2 potential cleanups and 1 clarification that nothing can be overwritten with the same filename pattern.

Snakefile
mumichae: I usually add the script as input, which is convenient when we update the script. I.e. every time the timestamp of a script is changed, the step would be rerun.

With larger input, we might want to rethink this, as every update to the pipeline would force to rerun the pipeline, which we might not want. But then, we'll need to have a good overview which of the scripts are updated and need to be rerun.

scripts/save_embeddings.py
```diff
@@ -0,0 +1,87 @@
...
+    else:
+        print('Calculating UMAP...')
+        sc.tl.umap(adata)
+        basis = 'umap'
+        label = 'UMAP'
+
+    # Save embedding plots
+    outdir = os.path.dirname(outfile)
```
mumichae, line 67: I'd require the file as input to the script instead of matching it to the output CSV file. Then snakemake can recompute when one of the images is missing. Otherwise we could run into the issue, that csv and plot don't match, e.g. when the script fails

scripts/snakemake_parse.py
mumichae: I agree that this workflow to determine the filenames is pretty clumsy. I'd suggest a complete refactoring of the pipeline based on my better understanding of snakemake now. This would make everything a lot easier.
But for now we should focus on getting things to run first.

mumichae: The files shouldn't be overwritten, as they are in different directories. Otherwise, snakemake will detect a conflict if multiple rules can create the same file. So nothing to worry about here.

mumichae (changes requested):
This seems pretty much done. I would only change the script to use a plot file name instead of creating it in the script. The integration into snakemake looks great and a dryrun worked for me.

With all the rules, it has become pretty impractical and difficult to understand. For the future, I would aim to refactor the file handling in the ParsedConfig class.

But for now, as long as it runs on our data, it should be fine.

Snakefile
LuckyMD: Typically, if we update a script we would be aware of what we're doing and therefore rerun the pipeline again as a result of that anyway I would imagine... otherwise why are you updating the script in the first place. Not sure you need an automated system taking care of the rerun there.

lazappi:
> Do we want the `all` rule to create the embedded images? Or should this be a separate rule that you can call for post-processing?

Yeah I wasn't sure either. I'm thinking most users will want to look at them but maybe we want them to focus on the metrics first 🤷‍♂️? Plus it's a bunch more steps to run... Happy to take that line out if we don't want it but should probably document somewhere that it exists.

Snakefile
lazappi: I don't really have an opinion either way about what is more correct but my vote is leaving it this way for now for consistency with the other stages.

scripts/save_embeddings.py
lazappi: That is a much better idea. I will change it to do that.

scripts/save_embeddings.py
lazappi: Are you suggesting I have three output arguments for the script (1 for the CSV and 2 for the PNGs) or just have the output directory as an argument?

Would Snakemake be able to track things if I add extra outputs to the stage? I'm not sure what we have done for the other steps that produce multiple outputs?

scripts/save_embeddings.py
lazappi: Ok. I did that. I have realised though that because ForceAtlas2 isn't in the environment it will always use the FR algorithm so I can probably just assume that is the label. Alternatively, we could add FA2 to the environment. Thoughts?

scripts/save_embeddings.py
mumichae: Yes, exactly.  Just as you define a named list for input you can create a named list for output and access it via `output.outputname` in the `shell` directive
```
rule abc:
...
    output:
         output1 = "umap.png"
         output2 = "other_file.txt"
    shell:
         """
         script.py { input} -o {output.output1} ... etc.
         """
```

lazappi: Ok. How's that now? I think Snakemake should be able to track the PNG files. I did try deleting one and it wasn't recreated but I think that is because it's not a direct output of `all`?

scripts/save_embeddings.py
LuckyMD: I think we should probably add fa2 to the environment...

LuckyMD (approved):

LuckyMD:
I think this is fine for now... `all` now actually means run everything.. so at least that's consistent. If @mumichae is happy with the changes, you can merge. It looks good from my side!

LuckyMD:
Might need to resolve the conflicts that emerged due to the benchmark file merging PR though.

scripts/save_embeddings.py
mumichae: Yeah, that's a good fix for now. It doesn't recreate the png files, because they aren't requested as input in the in the embedding_integrated and embedding_unintegrated rules. I'll  keep that as a Todo for when I restructure the pipeline, as it would be a mess to enforce it now.

mumichae (approved):

scripts/save_embeddings.py
LuckyMD: But I do think this is now a good solution as is either way. Adding FA2 is a separate issue.

PR #172, opened 2020-09-24
added Seurat RPCA mode

LuckyMD:
Add Seurat alignment mode using RPCA to find anchors.

### DIFF
--- a/R/integration.R
+++ b/R/integration.R
@@ -45,6 +45,10 @@ runSeuratRPCA = function(data, batch, hvg=2000) {
 	  require(Seurat)
 	  batch_list = SplitObject(data, split.by = batch)
 
+	  batch_list <- lapply(X = batch_list, FUN = function(x) {
+	  	     RunPCA(x, features = hvg)
+	  })
+
 	  anchors = FindIntegrationAnchors(
 	          object.list = batch_list,
 	          anchor.features = hvg,
--- a/R/integration.R
+++ b/R/integration.R
@@ -45,14 +45,14 @@ runSeuratRPCA = function(data, batch, hvg=2000) {
 	  require(Seurat)
 	  batch_list = SplitObject(data, split.by = batch)
 
-          batch_list <- lapply(X = batch_list, FUN = function(x) {
-		     ScaleData(object = x)
-	  	     RunPCA(x, features = hvg)
+          features <- SelectIntegrationFeatures(batch_list)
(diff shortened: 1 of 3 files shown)


danielStrobl (changes requested):
This doesn't seem to work, I get this error:
```
Loading required package: Seurat
Error in PrepDR(object = object, features = features, verbose = verbose) : 
  Data has not been scaled. Please run ScaleData and retry
Calls: runSeuratRPCA ... RunPCA -> RunPCA.Seurat -> RunPCA -> RunPCA.Assay -> PrepDR
Execution halted
```

danielStrobl:
Unfortunately this still doesn't work, I now get this error:
```
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'x' in selecting a method for function 't': subscript out of bounds
```

R/integration.R
```diff
@@ -39,6 +40,46 @@ runSeurat = function(data, batch, hvg=2000) {
         	   verbose = T)
 	  return(integrated)
 }
+
+runSeuratRPCA = function(data, batch, hvg=2000) {
+	  require(Seurat)
+	  batch_list = SplitObject(data, split.by = batch)
+
+          features <- SelectIntegrationFeatures(batch_list)
```
LuckyMD, line 48: this we want to avoid. So can we just put `features <- hvg` or something here for the time being?

LuckyMD (commented):
we need to use our HVG features... maybe draft sth that might do that and we can get input from Jain on how this can be done.

R/integration.R
danielStrobl: ok, I'll try that. The current setupd doesn't work either way

danielStrobl:
ok, I got it working in some cases now.
For the full_feature cases, I get this error here:
```
Error in nn2(data = c(1.57559389046269, -1.48019171953529, 0.398490401872583,  : 
  Cannot find more nearest neighbours than there are points
Calls: runSeuratRPCA ... RunIntegration -> FindWeights -> NNHelper -> do.call -> nn2
```

LuckyMD:
Looks good now! The above error is something that is Seurat internal. We ran into that a couple of times. It has to do with not finding enough anchors and can apparently happen with certain dataset sizes. That's a run fail that we can't control, but have to accept. HVG will work better anyway, so i think it's fine.

danielStrobl (approved):

PR #174, opened 2020-09-25
added new methods names

martaint:


LuckyMD (changes requested):
there will also be a `seurat_rpca`, which just got merged. So that might need to be called Seurat v3 RPCA and the other one "Seurat v3 CCA".

martaint:
> there will also be a `seurat_rpca`, which just got merged. So that might need to be called Seurat v3 RPCA and the other one "Seurat v3 CCA".

I added it! for the metrics file, please remove the underscore in seurat_rpca, that will cause problems in my stringsplit. seuratrpca will work!

LuckyMD:
> I added it! for the metrics file, please remove the underscore in seurat_rpca, that will cause problems in my stringsplit. seuratrpca will work!

@danielStrobl Have you started the runs already? And if not, could you quickly change this? Maybe just use "SeuratRPCA"

danielStrobl:
No, I haven't. Ok I'll adapt that

LuckyMD (approved):

PR #179, opened 2020-09-29
Merge benchmarks into a single output file

lazappi:
This PR adds a script and Snakemake rule for merging all the `.benchmark` files produced by the main pipeline (when timing is turned on). It is not part of the main pipeline and needs to be run separately after that is complete. The output is a `benchmarks.csv` file in the same location as the `metrics.csv` file.

Briefly what the script does is:
1. Search the root output directory for files in an `integration/` direction ending in `.benchmark`
2. Read in these files and concatenate them
3. Save the result

The output file can then be easily used to compare the scalability of methods.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -196,7 +196,7 @@ rule metrics_single:
         i      = get_integrated_for_metrics,
         script = "scripts/metrics.py"
     output: cfg.get_filename_pattern("metrics", "single")
-    message: 
+    message:
         """
         Metrics {wildcards}
         output: {output}
@@ -207,7 +207,7 @@ rule metrics_single:
         organism  = lambda wildcards: cfg.get_from_scenario(wildcards.scenario, key="organism"),
         assay     = lambda wildcards: cfg.get_from_scenario(wildcards.scenario, key="assay"),
         hvgs      = lambda wildcards: cfg.get_feature_selection(wildcards.hvg),
-        cmd       = f"conda run -n {cfg.py_env} python"	
+        cmd       = f"conda run -n {cfg.py_env} python"
     shell:
         """
         {params.cmd} {input.script} -u {input.u} -i {input.i} -o {output} -m {wildcards.method} \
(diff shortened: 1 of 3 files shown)


LuckyMD:
I generally really like this idea, but I feel it could be better implemented by taking the `config.yaml` file as input to get where all the .benchmark files are stored. You could check the `update_timestamp.py` script for how to do that. Maybe @mumichae has further input here as well.

lazappi:
Yeah happy to hear suggestions about how best to incorporate it. I wasn't sure the best way to pass the file names around so just went for the hacky quick solution. Maybe an added complication is they only exist if timing is turned on but I guess that should be clear from the config as well.

LuckyMD:
The timing from the config is legacy stuff that we no longer use. The `.benchmark` files are always generated if the Snakemake pipeline is used.

lazappi:
Ah, ok. It might be good to clean up all the old timing stuff if we can, I found it a bit confusing trying to work out what was going on.

LuckyMD:
Agreed. Do you want to write an issue for that so that it's on the to-do list for after revisions?

mumichae:
Collecting the benchmark files is pretty simple. If you check out the Snakefile, which is basically modified python code, you can see the pattern I use to create benchmark  files. You can put that pattern  into a variable or function for cleaner reuse. The `scripts/snakemake_parser.py` has a `ParsedConfig` class where I organise all the components used in the `Snakefile`.

Based on this I'd recommend to build on this class and include any snakemake feature in there. Then you can create  and call `cfg.getBenchmarkFiles()` to retrieved  whichever  set of files you want. Based on input and script, we can then   plug it into a rule.


A bit off topic but might help understand  the  code: It would also be a good idea to move this python class within the module. The way we have it now is a bit messy. Another  option  would be to separate the snakemake pipeline from scib entirely into a new repo, but that's  discussion  for a different  topic.

LuckyMD:
You should just be able to use `f'{cfg.get_filename_pattern("integration", "single", "h5ad")}.benchmark'` and `f'{cfg.get_filename_pattern("integration", "single", "rds")}.benchmark'` as in the integration rules, no?

mumichae (commented):
Looks good so far. Have you checked that a dry run works?

lazappi:
It works. I think it just needs to be adjusted to take input from Snakemake rather than just search for files. At the moment the Snakemake stage doesn't have an input.

mumichae:
I looked into it, and from the construct we need some proper refactoring of the way files are determined. Seeing as it is not top priority for the paper right now and it works, as long as at leats one benchmark output exists, I would delegate this to later and fix it in the new snakemake repo.

mumichae (approved):

lazappi:
@LuckyMD Happy for me to merge?

LuckyMD:
yes!

PR #184, opened 2020-10-04
Fix finding optimal resolution for clustering

mumichae:
**Problem**: Clustering does not use optimised resolution but last resolution of search space.
**Solution**: Corrected typo that broke finding optimal score for clustering.

This affects all metrics and other operations reliant on optimised clustering. If we have conclusions based on the clustering-based metrics, we might want to re-evaluate the metrics with this bugfix, if it is still feasible.


LuckyMD (approved):

LuckyMD:
This is very problematic... We need to redo the NMI and ARI metrics for all benchmarking results...

@danielStrobl do you have any idea how long it would take to only run the NMI and ARI metrics again? 

@mumichae could you write a script to only run the NMI and ARI metrics again and update the metrics outputs in the relevant files?

mumichae:
Yes, I'll look into it tomorrow

PR #185, opened 2020-10-05
Recompute metrics branch

mumichae:
Due to the fix in PR #184, some metrics need to be recomputed.

Affected metrics are:
+ ARI
+ NMI
+ isolated labels


mumichae:
Still having issues with the actual execution of the pipeline...

mumichae:
pipeline is now working in dryrun and on dummy dataset

LuckyMD:
A few questions:
1. Do we really want to merge this, or just use this branch to recompute the metrics? It seems this is otherwise just quite a specific case that is then built into our metrics script and function.
2. Where is the output stored? It would be good if we can replace the metrics.csv output to include the new outputs from this run. That will probably not be possible via Snakemake as another rule already has metrics.csv as an output.

mumichae:
1. I agree that we probably don't want this merged in master
2. The output will be in a separate file called `metrics_recomp.csv`. We can then impute these scores into a copy of the old `metrics.csv` before plotting

LuckyMD:
That sounds good to me. I would ideally like to have all the individual metrics.csv results updated as well though (post-processing?) so we don't store incorrect benchmarking information. Could you make another script for that? We can then just run `update_timestamp.py` afterwards on everything.

LuckyMD:
@danielStrobl Could you rerun from this branch?

mumichae:
> That sounds good to me. I would ideally like to have all the individual metrics.csv results updated as well though (post-processing?) so we don't store incorrect benchmarking information. Could you make another script for that? We can then just run `update_timestamp.py` afterwards on everything.

This hack only changes the time of the metrics computation, not the integration. Do we time the metrics as well?

LuckyMD:
> This hack only changes the time of the metrics computation, not the integration. Do we time the metrics as well?

The script (aka "hack" ^^) just updates the timestamp of all files involved in the snakemake pipeline... preprocessed files, integration outputs, and metrics runs. So it should still work.

mumichae:
Oh, I meant the clustering  recompilation hack... It shouldn't affect the timing of integration and metrics are not timed afaik?

I think it is useful to update the old metrics script, but it doesn't have to be part of the pipeline.

LuckyMD:
I think we're talking about different things here... maybe we can quickly discuss this f2f? Or via chat ;).

scripts/metrics.py
```diff
@@ -213,12 +213,12 @@
...
         silhouette_ = False
-        nmi_ = True
-        ari_ = True
-        pcr_ = False
-        cell_cycle_ = False
+        nmi_ = False
+        ari_ = False
+        pcr_ = True
+        cell_cycle_ = True
         isolated_labels_ = True
```
LuckyMD, line 222: @mumichae i think you missed `isolated_labels_ = False` here, no?

mumichae: thanks, I forgot to change that

mumichae:
Using this fix to run PC regression metrics

LuckyMD:
Are you sure you wanted to commit an updated viz notebook to a branch we will not merge?

danielStrobl:
Didn't want to change my working branch as there's still stuff running from
that. But I'll also add it to master

LuckyMD:
@danielStrobl @mumichae is there anything in here worth adding to the main repo or `scib-pipeline`? I imagine recomputing a single metric or a subset of metrics is not a bad function to have in general.

mumichae:
I agree, that might be a nice feature  to add to the pipeline

LuckyMD:
I think in general having a snakemake rule where the user can define outputs themselves is maybe not a bad idea... I'm not sure I would integrate it with the rest of the pipeline though. Then we could just have a different output filename?

PR #186, opened 2020-10-05
Include LISI files during package installation

mumichae:
LISI files are not being copied during pip installation of `scIB`. By specifying the missing files in the `setup.py`, they will be copied. However, that does not mean that the C source code is recompiled.


LuckyMD (approved):
perfect! Thanks a lot!

LuckyMD:
If we at some point figure out we can make the files be recompiled as well if we install via `pip` that would be amazing... but i guess this might not be possible.

PR #187, opened 2020-10-08
Isolated label fix

mumichae:
Recomputed isolated label due to ineffective cluster optimisation.

The isolated label score now optimises cluster resolution by F1 score between cluster assignment and isolated cell label. This is done for each isolated cell label separately.

### DIFF
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -376,14 +376,16 @@ def score_isolated_label(adata, label_key, batch_key, cluster_key,
     def max_label_per_batch(adata, label_key, cluster_key, label, argmax=False):
         sub = adata.obs[adata.obs[label_key] == label].copy()
         label_counts = sub[cluster_key].value_counts()
+        print(label_counts)
         if argmax:
+            print(label_counts.argmax())
             return label_counts.index[label_counts.argmax()]
         return label_counts.max()
 
     
     if cluster:
         opt_louvain(adata_tmp, label_key, cluster_key, function=max_label_per_batch,
-                    label=label, verbose=False)
+                    label=label, verbose=False, inplace=True)
     
         largest_cluster = max_label_per_batch(adata_tmp, label_key, 
                                               cluster_key, label, argmax=True)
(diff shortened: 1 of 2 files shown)


scIB/metrics.py
```diff
@@ -361,35 +360,44 @@ def get_isolated_labels(adata, label_key, batch_key, cluster_key, n, verbose):
...
-    import sklearn.metrics as scm
     adata_tmp = adata.copy()
-    
-    # cluster optimizing over cluster with largest number of isolated label per batch
+
     def max_label_per_batch(adata, label_key, cluster_key, label, argmax=False):
+        """cluster optimizing over cluster with largest number of isolated label per batch"""
         sub = adata.obs[adata.obs[label_key] == label].copy()
         label_counts = sub[cluster_key].value_counts()
+        print(label_counts)
```
LuckyMD, line 376: this looks like it shouldn't be turned on by default but only if verbose or sth, no?

scIB/metrics.py
```diff
@@ -4,6 +4,7 @@
 import pandas as pd
 import matplotlib.pyplot as plt
 import seaborn as sns
+import sklearn
```
LuckyMD, line 7: do you need this? Or does sklearn.metrics suffice?

LuckyMD (approved):
looks good overall.. just 2 small things.

scIB/metrics.py
mumichae: Yep, removed those

scIB/metrics.py
mumichae: We also want to use `sklearn.linear_model` for the PCR score, so I'd keep it the way it is. Also, it's nicer to see `sklearn.metrics` rather than `metrics` so we know exactly where that function comes from. I removed all imports from within functions, as they would only cause hidden errors

LuckyMD: okay

danielStrobl: this doesn't work. sklearn doesn't automatically import subpackages. I think you need an `import sklearn.metrics` there as well

PR #188, opened 2020-10-09
Fix LISI error for small values

mbuttner:
I replace some values in the connectivities matrix by the minimum value allowed as double in C++ (i.e. `1.7e-308`). Code is tested on two examples, however, I was unable to reproduce Luke's error message, when I set some values in the connectivities matrix to values smaller than `1.7e-308`.

### DIFF
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -1376,10 +1376,10 @@ def lisi_graph_py(adata, batch_key, n_neighbors = 90, perplexity=None, subsample
     if verbose:
         print("Compute knn on shortest paths") 
     
-    #set connectivities to 2e-305 if they are lower than 1.7e-308 (because cpp can't handle double values smaller than that).
+    #set connectivities to 3e-308 if they are lower than 1.7e-308 (because cpp can't handle double values smaller than that).
     connectivities = adata.uns['neighbors']['connectivities'] #csr matrix format
-    large_enough = connectivities.data>=2e-305
-    connectivities.data[large_enough==False] = 2e-305
+    large_enough = connectivities.data>=1.7e-308
+    connectivities.data[large_enough==False] = 3e-308
     
     #define number of chunks
     n_chunks = 1
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -1,4 +1,5 @@
 import pandas as pd
+import numpy as np
 import matplotlib.pyplot as plt
 import seaborn as sns
 #import networkx as nx
@@ -1379,6 +1380,11 @@ def lisi_graph_py(adata, batch_key, n_neighbors = 90, perplexity=None, subsample
     #set connectivities to 3e-308 if they are lower than 1.7e-308 (because cpp can't handle double values smaller than that).
     connectivities = adata.uns['neighbors']['connectivities'] #csr matrix format
     large_enough = connectivities.data>=1.7e-308
+    if verbose:
+        n_too_small = np.sum(large_enough==False)
+        if n_too_small:
+            print(f"{n_too_small} connectivities are too small and will be set to 3e-308")
+            print(connectivities.data[large_enough==False])
     connectivities.data[large_enough==False] = 3e-308
     
     #define number of chunks
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -1375,10 +1375,10 @@ def lisi_graph_py(adata, batch_key, n_neighbors = 90, perplexity=None, subsample
     if verbose:
         print("Compute knn on shortest paths") 
     
-    #set connectivities to 1.7e-308 if they are lower than 1.7e-308 (because cpp can't handle double values smaller than that).
+    #set connectivities to 2e-308 if they are lower than 1.7e-308 (because cpp can't handle double values smaller than that).
     connectivities = adata.uns['neighbors']['connectivities'] #csr matrix format
     large_enough = connectivities.data>=1.7e-308
-    connectivities.data[large_enough==False] = 1.7e-308
+    connectivities.data[large_enough==False] = 2e-308
     
     #define number of chunks
     n_chunks = 1
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -1375,10 +1375,10 @@ def lisi_graph_py(adata, batch_key, n_neighbors = 90, perplexity=None, subsample
     if verbose:
         print("Compute knn on shortest paths") 
     
-    #set connectivities to 2e-308 if they are lower than 1.7e-308 (because cpp can't handle double values smaller than that).
+    #set connectivities to 2e-305 if they are lower than 1.7e-308 (because cpp can't handle double values smaller than that).
     connectivities = adata.uns['neighbors']['connectivities'] #csr matrix format
-    large_enough = connectivities.data>=1.7e-308
-    connectivities.data[large_enough==False] = 2e-308
+    large_enough = connectivities.data>=2e-305
+    connectivities.data[large_enough==False] = 2e-305
     
     #define number of chunks
     n_chunks = 1
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -15,7 +15,7 @@ import cProfile
 from pstats import Stats
 import memory_profiler
 import itertools
-import multiprocessing
+import multiprocessing as multipro
 import subprocess
 import tempfile
 import pathlib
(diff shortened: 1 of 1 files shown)


LuckyMD (approved):
looks good to me. I like that you're only changing the input to the cpp part, and not the stored value. Is this tested?

LuckyMD:
@lazappi could you test this solves your problem?

lazappi:
Trying to test it but running into another error which is probably unrelated:

```python
/Users/luke.zappia/miniconda3/envs/scIB-python/lib/python3.7/site-packages/anndata/_core/anndata.py:21: FutureWarning: pandas.core.index is deprecated and will be removed in a future version.  The public classes are available in the top-level namespace.
  from pandas.core.index import RangeIndex
Trying to set attribute `.obs` of view, copying.
Trying to set attribute `.obs` of view, copying.
Trying to set attribute `.obs` of view, copying.
Trying to set attribute `.obs` of view, copying.
Traceback (most recent call last):
  File "scripts/metrics.py", line 248, in <module>
    trajectory_=trajectory_
  File "/Users/luke.zappia/Documents/Code/GitHub-theislab/scib/scIB/metrics.py", line 1888, in metrics
    multiprocessing = True, verbose=verbose)
  File "/Users/luke.zappia/Documents/Code/GitHub-theislab/scib/scIB/metrics.py", line 1504, in lisi_graph
    multiprocessing = multiprocessing, nodes = nodes, verbose=verbose)
  File "/Users/luke.zappia/Documents/Code/GitHub-theislab/scib/scIB/metrics.py", line 1390, in lisi_graph_py
    n_cpu = multiprocessing.cpu_count()
AttributeError: 'bool' object has no attribute 'cpu_count'
```

No error if I do it manually:

```python
>>> import multiprocessing
>>> multiprocessing.cpu_count()
12
```

I'll see if I can sort it out unless there are any suggestions.

Side note. The number of cores isn't passed from Snakemake so the metrics uses all of them which is a pet hate.

LuckyMD:
This might be due to refactoring of imports done by @mumichae yesterday.

mbuttner:
Phew, sometimes I wonder how the code has ever run properly. The issue is unrelated, actually. It's rather bad coding style, because I `import multiprocessing` and use a variable named `multiprocessing`. 
I changed the `import multiprocessing` to `import multiprocessing as multipro` and changed the respective calls in commit [5ac4642](https://github.com/theislab/scib/commit/5ac4642ffa9355b2396b81da59321c5f2ca8d002).

LuckyMD:
did you change the variable name `multiprocessing` as well? That would be the improvement of the coding style that would be nice ;).

lazappi:
Ok, I'm glad I hadn't completely messed something up. After fixing the multiprocessing thing I still get the original error:

```python
Trying to set attribute `.obs` of view, copying.
Trying to set attribute `.obs` of view, copying.
Trying to set attribute `.obs` of view, copying.
Trying to set attribute `.obs` of view, copying.
malformed matrix line 83 2661 1.700000000000000e-308
multiprocessing.pool.RemoteTraceback:
"""
Traceback (most recent call last):
  File "/Users/luke.zappia/miniconda3/envs/scIB-python/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/Users/luke.zappia/miniconda3/envs/scIB-python/lib/python3.7/multiprocessing/pool.py", line 47, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
  File "/Users/luke.zappia/Documents/Code/GitHub-theislab/scib/scIB/metrics.py", line 1263, in compute_simpson_index_graph
    if stat(input_path + '_indices_'+ str(chunk_no) + '.txt').st_size == 0:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/lisi_tmp1602242516/_indices_9.txt'
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "scripts/metrics.py", line 248, in <module>
    trajectory_=trajectory_
  File "/Users/luke.zappia/Documents/Code/GitHub-theislab/scib/scIB/metrics.py", line 1889, in metrics
    multiprocessing = True, verbose=verbose)
  File "/Users/luke.zappia/Documents/Code/GitHub-theislab/scib/scIB/metrics.py", line 1505, in lisi_graph
    multiprocessing = multiprocessing, nodes = nodes, verbose=verbose)
  File "/Users/luke.zappia/Documents/Code/GitHub-theislab/scib/scIB/metrics.py", line 1438, in lisi_graph_py
    count))
  File "/Users/luke.zappia/miniconda3/envs/scIB-python/lib/python3.7/multiprocessing/pool.py", line 276, in starmap
    return self._map_async(func, iterable, starmapstar, chunksize).get()
  File "/Users/luke.zappia/miniconda3/envs/scIB-python/lib/python3.7/multiprocessing/pool.py", line 657, in get
    raise self._value
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/lisi_tmp1602242516/_indices_9.txt'
```

The value is exactly `1.700000000000000e-308` now though so I think your replacing worked. Maybe it should not be exactly the smallest value?

mbuttner:
Reasonable. Changed it to `2e-308` in commit [09af8fb](https://github.com/theislab/scib/commit/09af8fb1bad8eb02309fad3a548143b7b01e7913).

lazappi:
Still no luck 😿 

```python
Trying to set attribute `.obs` of view, copying.
Trying to set attribute `.obs` of view, copying.
Trying to set attribute `.obs` of view, copying.
Trying to set attribute `.obs` of view, copying.
malformed matrix line 83 2661 2.000000000000000e-308
multiprocessing.pool.RemoteTraceback:
"""
Traceback (most recent call last):
  File "/Users/luke.zappia/miniconda3/envs/scIB-python/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/Users/luke.zappia/miniconda3/envs/scIB-python/lib/python3.7/multiprocessing/pool.py", line 47, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
  File "/Users/luke.zappia/Documents/Code/GitHub-theislab/scib/scIB/metrics.py", line 1263, in compute_simpson_index_graph
    if stat(input_path + '_indices_'+ str(chunk_no) + '.txt').st_size == 0:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/lisi_tmp1602244590/_indices_7.txt'
"""
```

lazappi:
I got this working with setting small values to 3e-308. I also added a print message about values being replaced.

I'm wondering if we should also change the threshold for small values to 3e-308 instead of 1.7e-308?

mbuttner:
> I got this working with setting small values to 3e-308. I also added a print message about values being replaced.

Awesome! Thank you for taking the time and looking into this!

> I'm wondering if we should also change the threshold for small values to 3e-308 instead of 1.7e-308?

Yes, that would be way more consistent.

LuckyMD:
Is this ready to be merged?

mbuttner:
Yes, I think so.

scIB/metrics.py
```diff
@@ -1375,14 +1377,24 @@ def lisi_graph_py(adata, batch_key, n_neighbors = 90, perplexity=None, subsample
...
     
+    #set connectivities to 3e-308 if they are lower than 3e-308 (because cpp can't handle double values smaller than that).
+    connectivities = adata.uns['neighbors']['connectivities'] #csr matrix format
+    large_enough = connectivities.data>=3e-308
+    if verbose:
+        n_too_small = np.sum(large_enough==False)
+        if n_too_small:
+            print(f"{n_too_small} connectivities are smaller than 3e-308 and will be set to 3e-308")
+            print(connectivities.data[large_enough==False])
+    connectivities.data[large_enough==False] = 3e-308
```
LuckyMD, line 1388: What happens to connectivity values that are 0 here? Are they also set to 3e-308?

LuckyMD (commented):
What happens to 0 connectivity values in this replacement?

scIB/metrics.py
mbuttner: This is a csr matrix. There are no explicit zero values in the matrix.

LuckyMD: ah... that's because you use `connectivities.data` which only takes non-zero values. If you just use the csr matrix directly, then you replace all the 0s... I just tried ;) :D

LuckyMD (approved):

PR #193, opened 2020-10-16
Handle empty input files when saving embedding plots

lazappi:
This was a bit trickier than I thought because I think the pipeline will complain if there is no output. What I have done is:

1. Check if the input `.h5ad` file is empty, if yes:
1. Delete any existing embedding output files
1. Create empty output files

I think it works but I have only tested it very quickly so probably good if @danielStrobl can try running it before we merge.

### DIFF
--- a/scripts/save_embeddings.py
+++ b/scripts/save_embeddings.py
@@ -5,7 +5,10 @@ import scanpy as sc
 import scIB
 import argparse
 import os
+import sys
 import warnings
+from pathlib import Path
+
 warnings.filterwarnings('ignore')
 
 RESULT_TYPES = [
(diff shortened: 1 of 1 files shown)


LuckyMD:
Why delete the existing embedding output files? What should there be if there is an empty h5ad file?

lazappi:
I assumed an empty `.h5ad` meant a failed run and that any existing embedding output was from some previous run so should be deleted. They will just get overwritten will blank files otherwise which is pretty much equivalent. Unless there is some other reason for an empty `.h5ad`. Easy enough to change if we want.

LuckyMD:
I might mean we were not careful in how we save our data if there is sth to delete. Did you come across this case? I would maybe throw an error if this happens.. what do you think? Typical Snakemake procedure would be to overwrite though.

lazappi:
Ah, I remember why I did this now. The `Path().touch()` command only changes the timestamp if the file exists so I had to delete it so out of date stuff was left there. I can probably come up with another way to overwrite with a empty file though.

Does anyone know what we do with the metrics if the `.h5ad` is empty? Guess we should probably do the same thing here.

LuckyMD:
In the metrics case we just output a file that has no value for all metric outputs (I think it was this, and not NAs). That is then captured in the metrics aggregation script to say the method wasn't run. @mumichae might be able to confirm this.

I think the safest approach here would be to throw an error if the file exists. Then we can check manually and just run the snakemake command again if needed.

mumichae:
In case the adata is empty, we would want empty output. If we had a non-empty output on the previous run, but for some reason the current run produces an empty adata, we wouldn't want to keep the previous embedding, so I'd overwrite any file that is output from the script but not captured by snakemake. But seeing as we provide all output filenames to the snakemake rule, snakemake should remove these files if they exist before a rerun. So all that you'll need to do is fix your script to output empty files.

mumichae:
So to clarify, when we have failed runs, they result in empty .h5ad files. That means an output file will be created. Snakemake doesn't execute a rule if there is no input for it. If you require snakemake to create an output file that it can't create, it'll throw a missing input error.

lazappi:
Ok, I changed how empty output files are created. I think this is more what we want but if not I'm probably a bit confused about what to do.

LuckyMD (approved):
I would say: yes... would you agree @mumichae ?

mumichae:
Yes, this looks fine

mumichae (approved):

Issue #192, opened 2020-10-12
'ncol(x)' is not equal to 'length(clusters)'


scottgigante:
Can't quite work out why this is happening, but I'm getting an error every time I normalize this dataset.

```
>>> import scIB
>>> import openproblems
>>> adata = openproblems.tasks.multimodal_data_integration.datasets.citeseq_cbmc(test=True)
>>> scIB.preprocessing.normalize(adata)
/home/scottgigante/.local/lib/python3.6/site-packages/scanpy/neighbors/__init__.py:121: FutureWarning: This location for 'distances' is deprecated. It has been moved to .obsp[distances], and will not be accesible here in a future version of anndata.
  adata.uns['neighbors']['distances'] = neighbors.distances
/home/scottgigante/.local/lib/python3.6/site-packages/scanpy/neighbors/__init__.py:122: FutureWarning: This location for 'connectivities' is deprecated. It has been moved to .obsp[connectivities], and will not be accesible here in a future version of anndata.
  adata.uns['neighbors']['connectivities'] = neighbors.connectivities
/home/scottgigante/.local/lib/python3.6/site-packages/scanpy/tools/_louvain.py:107: FutureWarning: This location for 'connectivities' is deprecated. It has been moved to .obsp[connectivities], and will not be accesible here in a future version of anndata.
  adjacency = adata.uns['neighbors']['connectivities']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/scottgigante/.local/lib/python3.6/site-packages/scIB/preprocessing.py", line 170, in normalize
    size_factors = ro.r(f'computeSumFactors(data_mat, clusters = input_groups, min.mean = {min_mean})')
  File "/home/scottgigante/.local/lib/python3.6/site-packages/rpy2/robjects/__init__.py", line 416, in __call__
    res = self.eval(p)
  File "/home/scottgigante/.local/lib/python3.6/site-packages/rpy2/robjects/functions.py", line 198, in __call__
    .__call__(*args, **kwargs))
  File "/home/scottgigante/.local/lib/python3.6/site-packages/rpy2/robjects/functions.py", line 125, in __call__
    res = super(Function, self).__call__(*new_args, **new_kwargs)
  File "/home/scottgigante/.local/lib/python3.6/site-packages/rpy2/rinterface_lib/conversion.py", line 44, in _
    cdata = function(*args, **kwargs)
[...]

PR #194, opened 2020-10-19
Added error checks for scran normalization

LuckyMD:
This PR addresses #192. Scran pooling normalization now checks for 0 count cell or gene input.


LuckyMD:
To do:
- [ ] test on small dataset
- [ ] test error cases

PR #197, opened 2020-10-22
Bugfix in PCA regression

mumichae:
+ fixed PCA Var instead of Var^2
+ proper order of data/response in regression
+ added tests


LuckyMD:
@LisaSikkema could you verify that this produces the same output as your code?

LuckyMD:
@LisaSikkema could you verify that this produces the same output as your code?

LuckyMD:
@mumichae did you intend to commit all your test scripts already here? Or is this for the travis branch?

LuckyMD (approved):
This looks addressed, but needs testing.

mumichae:
I intended the tests. They are basically running versions of what we had before. Pipeline tests are not yet working with the environments though.

LisaSikkema:
yes corresponds to my output for categorical covariates.
One thing I forgot to say: if you're working with a continuous covariate, the way the code is written now, the continuous variable is still converted to dummies (so one dummy for every age, in my data). Not sure if that matters? I assume your batches are always categorical. 
For the continuous data, if I skip the dummy-step and convert the covariate vector to the right shape, the values also match up perfectly.

LuckyMD:
That matters... do you need to set the continuous covariate to be numerical somewhere?

LisaSikkema:
It was numerical already. I suspect that's simply the way pd.get_dummies behaves when you feed it a single numerical vector

LuckyMD:
> Pipeline tests are not yet working with the environments though.

Maybe remove those that are not yet working then... we shouldn't merge things that are not tested.

mumichae:
I have now added a case for continuous variables and tested against kBET. I get a small deviation 3e-5. Unfortunately I don't have a good batch variable with a larger score, so I'm not sure if that is just numerical noise.

scIB/metrics.py
```diff
@@ -710,14 +710,24 @@ def pc_regression(data, variable, pca_var=None, n_comps=50, svd_solver='arpack',
...
-    
-    # one-hot encode categorical values
-    variable = pd.get_dummies(variable).to_numpy()
-    
+
+    # handle categorical values
+    try:
+        categorical = variable.dtype.name == 'category'
+    except:
+        categorical = not isinstance(variable[0], (int, float))
```
LuckyMD, line 722: Why use error handling and not if else?

scIB/metrics.py
```diff
@@ -710,14 +710,24 @@ def pc_regression(data, variable, pca_var=None, n_comps=50, svd_solver='arpack',
...
+
+    if categorical:
+        if verbose:
+            print("one-hot encode categorical values")
+        variable = pd.get_dummies(variable).to_numpy()
+    else:
+        variable = np.array(variable).reshape(-1, 1)
```
LuckyMD, line 729: why do you reshape here?

LuckyMD (commented):
please use if/else and not try/except for this.

scIB/metrics.py
mumichae: You mean check if the variable is of type series/dataframe with  an if/else? I found  this approach to be more readable and general towards whether it congestions a dtype or not.  Would you prefer if/else type checking instead?

scIB/metrics.py
mumichae: To make it compatible with lm. I would otherwise get an error. Any alternative suggestions?

mumichae:
We're dealing with potentially different datatypes here. I don't think it would make that much of a difference whether we use an exception or an if, it would rather be a cosmetic change.

scIB/metrics.py
LuckyMD: yes. I don't think we should catch all errors like this when we know what we are looking for. Ideally check if categorical, check if numerical... if neither throw an error.

scIB/metrics.py
LuckyMD: All good... i thought a transpose or sth should work as well, but that doesn't work for me ^^.

LuckyMD:
try/except catches all errors... that's a bit overkill... we should be specific and allow the code to throw errors if need be.

scIB/metrics.py
LuckyMD: Please do this as if else...

mumichae: I included a different  type check (with if/else) which will be in the  next commit, once the values of kBet are reproducible

LisaSikkema:
Yes, pr_regression bit works for me and corresponds to output of my own code.

mumichae:
When I tried to reproduce the values in kBET pcRegression, I get completely different values for the pbmc3k dataset with batches [scanpy tutorial](https://scanpy-tutorials.readthedocs.io/en/latest/integrating-data-using-ingest.html), [scIB pytest function](https://github.com/theislab/scib/blob/706a11531921ee079032f97dbe99e8dce97b2064/tests/conftest.py#L26)

The conversion of categorical values to dummy values might be the cause for this discrepancy. https://github.com/theislab/scib/blob/706a11531921ee079032f97dbe99e8dce97b2064/scIB/metrics.py#L724

@LisaSikkema What code are you comparing against?

LisaSikkema:
I'm comparing this code of yours (I wrote the first 5 lines to make my object compatible with your code):
```
cov = "age" # also tried "dataset"
verbose = True
X_pca = subadata.obsm["X_pca"]
pca_var = subadata.uns["pca"]["variance"]
n_comps = 50
variable = subadata.obs[cov].copy()
try:
    categorical = variable.dtype.name == "category"
except AttributeError:
    categorical = not isinstance(variable[0], (int, float))

if categorical:
    if verbose:
        print("one-hot encode categorical values")
    variable = pd.get_dummies(variable).to_numpy()
else:
    variable = np.array(variable).reshape(-1, 1)

# fit linear model for n_comps PCs
r2 = []
for i in range(n_comps):
    pc = X_pca[:, [i]]
    lm = sklearn.linear_model.LinearRegression()
    lm.fit(variable, pc)
    r2_score = lm.score(variable, pc)
    r2.append(r2_score)

Var = pca_var / sum(pca_var) * 100
R2Var = sum(r2 * Var) / 100
```

against this code of mine:
```
var_explained = pd.DataFrame(index=range(n_pcs), columns=covariates + ["overall"])
for pc in range(n_pcs):
    y_true_unfiltered = subadata.obsm["X_pca"][:, pc]
    var_explained.loc[pc, "overall"] = np.var(y_true_unfiltered)
    for cov in covariates:
        x = subadata.obs[cov].values.copy()
        x_nans = np.vectorize(utils.check_if_nan)(x)
        x = x[~x_nans]
        y_true = y_true_unfiltered[~x_nans].reshape(-1, 1)
        if x.dtype in ["float32", "float", "float64"]:
            x = x.reshape(-1, 1)
        else:
            if len(set(x)) == 1:
                var_explained.loc[pc, cov] = np.nan
                continue
            x = pd.get_dummies(x)
        lrf = LinearRegression(fit_intercept=True).fit(
            x,
            y_true,
        )
        y_pred = lrf.predict(x)
        var_explained.loc[pc, cov] = np.var(y_pred)
total_variance_explained = np.sum(var_explained, axis=0).sort_values(ascending=False)
total_variance_explained_fractions = (
    total_variance_explained / total_variance_explained["overall"]
)
```

for variables that have no nan or "nan" entries.

danielStrobl:
> When I tried to reproduce the values in kBET pcRegression, I get completely different values for the pbmc3k dataset with batches [scanpy tutorial](https://scanpy-tutorials.readthedocs.io/en/latest/integrating-data-using-ingest.html)

Hey! So I've just tried to reproduce that with the dataset generated by the `adata_batch()` function. I'm getting very similar scores for both your implementation and the kBET one ( `8.99e-05`). How did you test the kBET implementation?

LuckyMD (approved):

PR #201, opened 2020-10-30
change adata to adata_tmp in LISI_graph

mbuttner:


LuckyMD:
Can we test on snakemake using only pancreas (but all metrics) to see if this changes lisi scores?

LuckyMD (approved):
simple fix, just requires testing if this addresses the bug we found.

danielStrobl:
Screenshot from 2020-10-30 17-31-52
Seems like this doesn't affect our scores, so we don't need to rerun.

danielStrobl (approved):

LuckyMD:
Nice! So this is probably only an issue for subsetted objects then @Hrovatin! We will merge but will then not rerun fully.

PR #202, opened 2020-11-09
Deactivate anndata2ri after preprocessing

scottgigante:
Best practice is probably to use local converters instead. 

https://github.com/theislab/anndata2ri/issues/50


LuckyMD (approved):

LuckyMD:
Thx for the contribution. To be honest... we don't even need `anndata2ri` here, but the other numpy conversion should be enough. For another time...

PR #203, opened 2020-11-10
Add option to use leiden

scottgigante:
Louvain is buggy and not recommended, as per https://github.com/theislab/scanpy/issues/1191#issuecomment-622229246

### DIFF
--- a/scIB/preprocessing.py
+++ b/scIB/preprocessing.py
@@ -143,7 +143,7 @@ def plot_count_filter(adata, obs_col='n_counts', bins=60, lower=0, upper=np.inf,
         plt.show()
 
 ### Normalisation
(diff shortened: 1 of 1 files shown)


LuckyMD:
Hmm... Louvain shouldn't really have changed since 2008. We could just pin it to version <0.7 (I still use 0.6.1 I think). It doesn't really matter though... leiden and louvain shouldn't matter much for knn graphs.

LuckyMD:
There are quite a lot more places where louvain is used in the metrics. So if we change this, we should change it everywhere. I'd prefer to do this after the revisions are handed in though. How urgent is this for openproblems?

LuckyMD:
@danielStrobl can you confirm the version you are currently using in the environment?

PR #205, opened 2020-12-03
GitHub actions integration

mumichae:
+ using pip environment without ML integration methods
+ only metrics tests included, not integration runs or R-based methods (yet)
+ general metric checks (mostly unmodified)
+ using `scanpy` PBMC datasets with 2 batches (except for cell cycle score)


LuckyMD:
This should only be merged when the `scIB-python` PR #147 is merged.... @danielStrobl ;)

mumichae:
@LuckyMD Could we prioritise this PR over the scanpy version PR? It would be great to update the dependencies on GA directly. The only tests missing here are kBET and LISI scores, but we have a solid requirements list for scib only and running tests for all other metrics. kBET and LISI could be addressed instead in the current scanpy update PR or a new PR.

LuckyMD:
Yes, let's do kBET and LISI in a new PR... let me just review quickly.

tests/conftest.py
```diff
@@ -1,6 +1,72 @@
...
+def adata_pca():
+    def adata_pca_(adata):
+        scIB.pp.reduce_data(adata, pca=True, n_top_genes=200, neighbors=False, umap=False)
+        return adata
+    return adata_pca_
+
+
+@pytest.fixture(scope="session")
+def adata_neighbors():
+    def adata_neighbors_(adata):
```
LuckyMD, line 55: here also strange inner function

tests/conftest.py
```diff
@@ -1,6 +1,72 @@
...
+        scIB.pp.reduce_data(adata, pca=True, n_top_genes=200, neighbors=True, umap=False)
+        return adata
+    return adata_neighbors_
+
+
+@pytest.fixture(scope="session")
+def adata_clustered():
+    def adata_clustered_(adata):
```
LuckyMD, line 63: again... not sure i understand this...

tests/metrics/test_pcr_metrics.py
```diff
@@ -0,0 +1,39 @@
...
+def test_pcr_batch(adata):
+    # no PCA precomputed
+    score = scIB.me.pcr_comparison(
+        adata, adata,
+        variable='batch',
+        n_comps=50,
+        scale=True
+    )
+    LOGGER.info(f"no PCA precomputed: {score}")
+    assert 0 <= score <= 1 and score < 1e-6
```
LuckyMD, line 18: bit strange this one... `score <= 1` and `score < 1e-6`... first one seems redundant.

tests/metrics/test_pcr_metrics.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_pcr_batch_precomputed(adata, adata_pca):
+    adata = adata_pca(adata)
+    score = scIB.me.pcr_comparison(adata, adata, variable='batch', scale=True)
+    LOGGER.info(f"precomputed PCA: {score}")
+    assert 0 <= score <= 1 and score < 1e-6
```
LuckyMD, line 25: <=1 redundant again.

tests/metrics/test_pcr_metrics.py
```diff
@@ -0,0 +1,39 @@
...
+    score = scIB.me.pcr_comparison(
+        adata_pre=adata,
+        adata_post=embed_factory(adata, type_='full'),
+        variable='batch',
+        embed='X_emb',
+        n_comps=50,
+        scale=True
+    )
+    LOGGER.info(f"using embedding: {score}")
+    assert 0 <= score <= 1 and score < 1e-6
```
LuckyMD, line 39: also remove <=1

LuckyMD (changes requested):

LuckyMD:
If I saw correctly, you are removing only the cProfiler and R func_profiler test files, right? Those were old so not needed. Anything else you are removing? I notice there were a lot of duplicate files. Also, I'm not sure why you are changing metrics code in this PR. Seems like something for another PR.

envs/scIB-R-conversion.yml
```diff
@@ -4,6 +4,7 @@ channels:
   - bioconda
   - conda-forge
 dependencies:
+  - r-optparse
   - _libgcc_mutex=0.1=main
   - _r-mutex=1.0.0=anacondar_1
   - anndata2ri=1.0.2=py_0
```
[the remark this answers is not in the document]
mumichae: I don't handle this here, but in the `scib-pipeline` repo. So all changes on the `envs` files can be ignored

requirements.txt
```diff
@@ -0,0 +1,32 @@
...
+python-igraph
+llvmlite==0.30
+memory_profiler
+
+# integration tools
+bbknn==1.3.9
+scanorama==1.7.0
+mnnpy==0.1.9.5
+scgen==1.1.5
+scvi==0.6.7
```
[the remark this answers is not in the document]
mumichae: This file only handles pip requirements. No R methods can be specified here.

scIB/metrics.py
```diff
@@ -558,98 +561,62 @@ def cell_cycle(adata_pre, adata_post, batch_key, embed=None, agg_func=np.mean,
...
     scores_before = []
     scores_after = []
-    #if both (s-score, g2m-score) and pc-regression are pre-computed 
-    if (np.in1d(['S_score', 'G2M_score'], 
-                adata_pre.obs_keys()).sum() == 2) and ('scores_before' in adata_pre.uns_keys()): 
-        #extract needed infos from adata_pre and delete it from memory
-        df_pre = adata_pre.obs[['S_score', 'G2M_score', batch_key]]
-        
-        scores_precomp = pd.Series(adata_pre.uns['scores_before'])
-        del adata_pre
```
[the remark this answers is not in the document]
mumichae: There was duplicated code here for some reason. We could move that to a separate PR though.

tests/conftest.py
```diff
@@ -1,6 +1,72 @@
...
+
+@pytest.fixture(scope="function")
+def adata(adata_pbmc):
+    adata = adata_pbmc.copy()
+    yield adata
+    del adata
+
+@pytest.fixture(scope="session")
+def adata_pca():
+    def adata_pca_(adata):
```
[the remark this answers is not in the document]
mumichae: This is a fixture factory. That way, separate copies of objects (known as fixtures in pytest) can be created.
https://docs.pytest.org/en/stable/fixture.html#factories-as-fixtures

scIB/metrics.py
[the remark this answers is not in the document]
mumichae: Reverted these changes and moved the, to #219

mumichae:
I didn't remove too much. I had to undo some changes, since this PR is a bit old already and some of the issues have already been taken care of in other PRs and the `scib-pipeline`. Also, I moved the metrics changes to #219 

The inner functions in the pytest setup are fixture factories that enable a new instance of objects for each metric as a copy of a longer-living object. As a result

1. I reduce the runtime, because no new objects need to be created from scratch but are copied instead
2. I ensure that each function has its own copy of the dataset and is not influenced by any other metric run on the same object

tests/conftest.py
```diff
@@ -1,6 +1,72 @@
...
+
+    yield adata_concat
+    del adata_concat
+
+
+@pytest.fixture(scope="function")
+def adata(adata_pbmc):
+    adata = adata_pbmc.copy()
+    yield adata
+    del adata
```
LuckyMD, line 43: `scope=function` is redundant here, right? But good to leave I guess as it is more explicit. The scope ensures that a different copy is received by every test function if I understand correctly?

LuckyMD: Naming here is not really great... as this fixture can be confused with the object name. Maybe rename to `adata_pbmc_factory` or so.

tests/metrics/test_beyond_label_metrics.py
```diff
@@ -0,0 +1,53 @@
+from tests.common import *
+
+
+def test_cell_cycle(adata_factory):
+    adata = adata_factory()
```
LuckyMD, line 5: any reason why `adata_factory` here, but `adata` below? Here you always re-execute the same function to generate a new `adata` object. It might be worth making `adata_factory` a regular fixture, rather than a factory. Any reason why you want to re-execute that code?

tests/conftest.py
```diff
@@ -1,6 +1,72 @@
...
+@pytest.fixture(scope="session")
+def adata_pca():
+    def adata_pca_(adata):
+        scIB.pp.reduce_data(adata, pca=True, n_top_genes=200, neighbors=False, umap=False)
+        return adata
+    return adata_pca_
+
+
+@pytest.fixture(scope="session")
+def adata_neighbors():
```
LuckyMD, line 54: is this intentionally not dependent on the `adata` fixture? You only ever use these together I think. It's fine to make this more general i guess, but then is order of fixtures given in the tests below?

LuckyMD: Never mind... you don't use these as dependent fixtures, but as separate calls.

LuckyMD (approved):
In general this is ready to be merged. Just 2 points:
1. `adata` fixture should be named a bit more standard so it's clear it is a fixture.
2. `adata_factory` use is a little confusing to me. Why re-execute the whole code to use it as a factory only for cell_cycle tests? Could make this a fixture. In general leave the 2 separate adata fixtures though, as it will help us in the long run.

tests/conftest.py
mumichae: I guess this could have been a simple function instead of a fixture, tbh

tests/metrics/test_beyond_label_metrics.py
mumichae: I agree. Originally I used `adata_factory` differently, but then I introduced the `adata_pbmc` factory. For cell cycle scores, the PBMC didn't result in enough cell cycle genes, which is why I had to use a different dataset here.

tests/conftest.py
mumichae: I changed it to a fixture

tests/metrics/test_beyond_label_metrics.py
mumichae: I changed this factory to a fixture called `adata_paul15`. Ideally, we would want a single adata fixture which would make this one obsolete. I'd keep it like this for now though.

tests/metrics/test_pcr_metrics.py
```diff
@@ -18,18 +18,17 @@ def test_pcr_batch(adata):
...
     assert 0 <= score < 1e-6
 
 
-def test_pcr_batch_embedding(adata, embed_factory):
+def test_pcr_batch_embedding(adata):
     # use different embedding
     score = scIB.me.pcr_comparison(
         adata_pre=adata,
-        adata_post=embed_factory(adata, type_='full'),
+        adata_post=add_embed(adata, type_='full'),
```
LuckyMD, line 31: as this is now a function, does this need to be imported here? Or do you want to keep this as a fixture factory?

tests/preprocessing/test_clustering.py
```diff
@@ -1,9 +1,10 @@
 from tests.common import *
 
 
-def test_cluster(adata, adata_neighbors):
+def test_cluster(adata):
+    scIB.pp.reduce_data(adata, pca=True, n_top_genes=200, neighbors=True, umap=False)
     _, _, score_all, clustering = scIB.cl.opt_louvain(
-        adata_neighbors(adata),
+        adata,
```
LuckyMD, line 7: why not just require `adata_neighbors` here? This seems redundant with the `adata_neighbors` fixture.

LuckyMD (changes requested):
Mainly use `adata_neighbors` more consistently.

tests/metrics/test_cluster_metrics.py
```diff
@@ -11,34 +11,38 @@ def test_ari_trivial(adata):
...
 
 
-def test_nmi(adata, adata_neighbors, cluster_factory):
-    _, _, nmi_all = cluster_factory(
-        adata_neighbors(adata),
-        cluster_key='cluster',
+def test_nmi(adata):
+
+    _, _, nmi_all = scIB.cl.opt_louvain(
+        adata,
```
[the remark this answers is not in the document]
mumichae: Thanks for spotting that, I must have overlooked this

tests/metrics/test_pcr_metrics.py
mumichae: The function is declared in `common.py` now, so it doesn't need to be imported explicitly for pytest to find it.

tests/preprocessing/test_clustering.py
mumichae: forgot about that one as well...

mumichae:
Thanks for the thorough review! I fixed the last few issues, so this branch should be ready to merge now

LuckyMD (approved):

PR #206, opened 2020-12-12
Analysis notebook updates

LuckyMD:
updated analysis notebooks for detailed metrics analysis and added correlation analysis notebook


LuckyMD:
Should I merge this here or transfer the PR to scib-reproducibility @lazappi? I guess you still have to transfer a lot of stuff across anyway, no?

lazappi:
If this is changing files probably better to do it here first? I don't think you can transfer a PR but I guess you could delete them here and just subject the new ones to `scib-reproducibility`. That would also work as long as you are able to keep track of things.

LuckyMD:
I think I'll go with first merging here, and then moving across then.

PR #208, opened 2021-01-19
Avoid filtering in normalization

scottgigante:
If cells are filtered in this stage, scran throws an error: 
`error in evaluating the argument 'object' in selecting a method for function 'sizeFactors': 'ncol(x)' is not equal to 'length(clusters)'`


LuckyMD:
Default should be `min_counts=1` if this is not kept you can't run total counts normalization as you'd be dividing by zero. I think the solution would be to ensure no 0 count cells are in the dataset. Can't normalize those anyway, no?

mumichae:
Shall we set `min_counts=1` then?

Issue #210, opened 2021-02-08
Small bug in silhouette_batch


alitinet:
Hi,

there's a small bug in `silhouette_batch()`: when there are only 2 cells (which is rarely the case but still) in `adata_group`, then `sklearn.metrics.silhouette_samples` crashes with 

```
ValueError: Number of labels is 2. Valid values are 2 to n_samples - 1 (inclusive),
```
even though it says that 2 is a valid value?

Thanks for the metrics, super useful.

Hrovatin:
I just got the same error when calculating cell type ASW. For me it happened because a sample passed to sklearn silhouette  has only 2 instances (e.g. a batch-cell type group with only two instances). 
For my problem here https://github.com/theislab/scib/blob/8f6c6524ef771af504f5a30629d83045becc0dc1/scIB/metrics.py#L77  a condition should be added, so that the check becomes:
```
if adata_group.obs[batch_key].nunique() == 1 or adata_group.shape[0]<3:
    continue
```
I think this should work.

LuckyMD:
Thanks @Hrovatin! Would you mind making a quick PR with this? Then you would get credit for the fix properly :).

Hrovatin:
see #212

PR #212, opened 2021-02-09
fix bug in silhouete calculation

Hrovatin:
Correct bug from #210


LuckyMD:
This is not an issue for other silhouette metrics? I guess it's very unlikely that this happens either way... so might be fine to ignore other cases for now.

LuckyMD (approved):

Hrovatin:
The code worked for me after this fix. But this could happen whenever n_classes==n_instances in sklearn silhouette (if it is used somewhere else as well - I did not check).

LuckyMD:
It's used in a couple more scores... but if someone has issues with that, we can move this check into another more central silhouette function :).

PR #213, opened 2021-02-16
Isolated label asw embed

mumichae:
## Issue
`isolated_label()` metric did not allow to specify which embedding to use, and uses `X_pca` instead.

## Solutions
- add `embed` option to `isolated_label` function
- require `embed` in `silhouette()` as a positional parameter
- add `use_rep` parameter for `opt_louvain()`


scIB/metrics.py
```diff
@@ -41,27 +41,34 @@ def __init__(self, message):
...
 
 ### Silhouette score
-def silhouette(adata, group_key, metric='euclidean', embed='X_pca', scale=True):
+def silhouette(adata, group_key, embed, metric='euclidean', scale=True):
     """
-    wrapper for sklearn silhouette function values range from [-1, 1] with 1 being an ideal fit, 0 indicating overlapping clusters and -1 indicating misclassified cells
+    wrapper for sklearn silhouette function values range from [-1, 1] with 1 being an ideal fit, 0 indicating
+    overlapping clusters and -1 indicating misclassified cells
+    :param group_key: key in adata.obs of cell labels
+    :param embed: embedding key in adata.obsm, usually 'X_pca'
```
LuckyMD, line 49: "default" instead of "usually"

scIB/metrics.py
```diff
@@ -351,22 +381,23 @@ def get_isolated_labels(adata, label_key, batch_key, cluster_key, n, verbose):
...
+        embed,
+        cluster=True,
+        cluster_key='iso_cluster',
+        verbose=False
+):
     """
     compute label score for a single label
     params:
         cluster: if True, use clustering approach, otherwise use silhouette score approach
+        kwargs:
```
LuckyMD, line 397: this was removed, so no longer need kwargs. Could you complete the function description here though?

scIB/metrics.py
```diff
@@ -302,14 +309,24 @@ def ari(adata, group1, group2):
...
+        verbose=True
+):
     """
     score how well labels of isolated labels are distiguished in the dataset by
         1. clustering-based approach
         2. silhouette score
     params:
         cluster: if True, use clustering approach, otherwise use silhouette score approach
+        embed: key in adata.obsm used for silhouette score if cluster=True or
+            as representation for clustering (if neigbors missing in adata)
```
LuckyMD, line 329: In our current pipeline reduce_data is always called though, so the F1 score never had an issue, right? The addition to neighbourhood graphs for cluster-based iso label scoring is a good fall-back, but it was never an issue, no?

scIB/metrics.py
```diff
@@ -41,27 +41,34 @@ def __init__(self, message):
...
+        metric=metric
+    )
     if scale:
         asw = (asw + 1)/2
     return asw
 
-def silhouette_batch(adata, batch_key, group_key, metric='euclidean', 
-                     embed='X_pca', verbose=True, scale=True):
+def silhouette_batch(adata, batch_key, group_key, embed, metric='euclidean',
+                     verbose=True, scale=True):
```
LuckyMD, line 64: To keep this clean, we need to adapt how this is called in the metrics function if you change the order of parameters.

LuckyMD (changes requested):
A few changes:
1. wording "default" instead of "usually"
2. Clean up order of parameters in function call to silhouette_batch
3. function description of score_isolated_label
4. Cluster_key is now always set to overwrite `adata.obs['iso_label']`. That's probably fine as this was done in the pipeline anyway. Not really an issue...

scIB/metrics.py
mumichae: The way we currently have it in the `metrics` function is 
```
sil_global = silhouette(adata_int, group_key=label_key, embed=embed, metric=si_metric)
```
So no changes needed.

scIB/metrics.py
```diff
@@ -384,11 +415,20 @@ def max_f1(adata, label_key, cluster_key, label, argmax=False):
         return max_f1
 
     if cluster:
-        opt_louvain(adata_tmp, label_key, cluster_key, function=max_f1, label=label, verbose=False, inplace=True)
+        opt_louvain(
+            adata_tmp,
+            label_key,
+            cluster_key,
+            use_rep=embed,
```
mumichae, line 422: The option `use_rep` could lead to different clustering results, which would change the isolated label score.

scIB/metrics.py
mumichae: Good point. The clustering for the F1 isolated label score is independent from the global clustering used by ARI, NMI etc., called by `reduce_data`. The latter case was using the correct embedding, even when `recompute_pca` was erroneously set to `True` for embedding output. However, the isolated label F1 score would have been computed on the `X_pca` instead of the actual embedding, so that we would expect different results.

mumichae: Correct, `use_rep` is only used if `adata.uns[neighbors]` is missing. In cases where `reduce_data` is not run, that'll ensure that the neighborhood graph is computed on the correct representation.

scIB/metrics.py
LuckyMD: This is for the cluster ASW, but not for the `silhouette_batch()` call if I'm not mistaken. There you call these arguments in a different order i believe.

mumichae: The arguments are defined and passed as `silhouette_batch(adata, batch_key, group_key, embed)`

LuckyMD: hmm, i seem to have misread. You're right.

LuckyMD (approved):

PR #215, opened 2021-02-28
Fix: Anndata concatenate causes shuffled batches

mumichae:
When we use `Anndata.concatenate` and specify the `batch_key` parameter, we need to ensure that the correct batch categories are passed to the method as well. Otherwise these will be overwritten with batches `0:n_batches-1`.
On the test data, the batch key is 'batch', the same as the default value of `batch_key` in `Anndata.concatenate`. This results in shuffled batches after concatenating anndata objects split by batch.

Fixes

- provide option in `utils.splitBatches` to return the batches categories
- pass batch categories to `Anndata.concatenate` function together with `batch_key`, so that correct batch annotations are retained

### DIFF
--- a/scIB/integration.py
+++ b/scIB/integration.py
@@ -6,20 +6,14 @@
     as well as tools and metrics to benchmark them.
 """
 
-import scanpy as sc
 import scipy as sp
-#import numpy as np
 from scIB.utils import *
-from memory_profiler import profile
 import os
-import pandas as pd
 import anndata
 
 import rpy2.rinterface_lib.callbacks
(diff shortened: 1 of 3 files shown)


LuckyMD (approved):
I added a small fix that also makes the `adata.obs_names` consistent to before concatenation. We were fixing this in a hacky way in the metrics.py script the whole time. This is a better approach though. Could you confirm this still works in your pipeline @mumichae and then we can merge?

mumichae:
The pipeline runs through :+1:

PR #218, opened 2021-03-10
[WIP] Remove scanpy<1.5 requirement

danielStrobl:
I tried to remove all references to `adata.uns['neighbors']['connectivities']|['distances']` here to make scIB compatible with newer versions of scanpy and anndata.
Still needs testing though.


LuckyMD:
Is this sth that is testable in your testing pipeline @mumichae ?

LuckyMD:
Also, @danielStrobl... this version now only works for scanpy >= 1.5. could you maybe add a check for neighbors being stored in `.obsp` maybe in the beginning of the metrics function? And then we'd need to reflect scanpy>=1.5 in the environment as well.

mumichae:
The pipeline seems to run through when I upgrade scanpy to version 1.7

danielStrobl:
> Also, @danielStrobl... this version now only works for scanpy >= 1.5. could you maybe add a check for neighbors being stored in `.obsp` maybe in the beginning of the metrics function? And then we'd need to reflect scanpy>=1.5 in the environment as well.

ok, I'll add that

mumichae:
@danielStrobl Are you able to set up the scib-pipeline test in your own environment and get it running?
From what I can see, `anndata==0.7.1` vs `anndata==0.7.5` is what causes the difference here.

LuckyMD:
Does your pipeline test also run the metrics? We would need to test that the metrics work as well. Also kBet, which is affected by this. With the new adaptation from @danielStrobl, maybe also on the old data as input (just test metrics rule).

mumichae:
Yes, the test is end-to-end. I didn't modify any code in the pipeline scripts and things ran through without error. Here are the results of scanpy 1.4.6, anndata 0.7.1 (master) vs. scanpy 1.7, anndata 0.7.5 (scanpy17)
scanpy17-master

LuckyMD:
This looks very good @mumichae! Thanks! Just changes on the stochastic methods (scVI, scANVI), which will deal less well with very small test data. 

As soon as the test for old vs new AnnData object is there, we can merge @danielStrobl !

danielStrobl:
I don't think we explicitly need to check for the correct location, I think scanpy/anndata do that for us.
This is what I get when I import an old file:
```
/home/daniel/.conda/envs/scop/lib/python3.7/site-packages/anndata/compat/__init__.py:182: FutureWarning: Moving element from .uns['neighbors']['distances'] to .obsp['distances'].

This is where adjacency matrices should go now.
```

LuckyMD:
> I think scanpy/anndata do that for us.

Awesome! Then this is ready for merge from my side. But I think we should give prio to #205. I will approve this for now though.

LuckyMD (approved):

LuckyMD:
First time GA is being used to check the PR... woop woop ;).

mumichae:
@danielStrobl @LuckyMD I can't seem to find where the scanpy and anndata versions are pinned. Do we need a new PR for that?

LuckyMD:
they're not pinned anymore with this PR...

PR #219, opened 2021-03-14
Remove redundant cell cycle code

mumichae:
For some reason, the cell cycle score contains redundant code, which makes it very difficult to debug.

Other issues addressed here:
- make it possible to enforce recomputation of cell cycle scores, even if they exist
- rename 'covariate' to 'variable' to avoid confusion


tests/metrics/test_pcr_metrics.py
```diff
@@ -0,0 +1,38 @@
+from tests.common import *
+import pandas as pd
+
+
+def test_pc_regression(adata):
+    scIB.me.pc_regression(adata.X, adata.obs["batch"])
```
mbuttner, line 6: What is tested here?

mbuttner (commented):
Looks good overall, I did not fully understand all `assert` calls. Is this tested?

tests/metrics/test_pcr_metrics.py
LuckyMD: implementation... if it doesn't run, there's an errorr

tests/metrics/test_silhouette_metrics.py
```diff
@@ -0,0 +1,40 @@
...
+    score = scIB.me.isolated_labels(
+        adata_pca,
+        label_key='celltype',
+        batch_key='batch',
+        embed='X_pca',
+        cluster=False,
+        verbose=True
+    )
+    LOGGER.info(f"score: {score}")
+    assert score <= 1
```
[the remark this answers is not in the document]
LuckyMD: just testing upper and lower boundaries are correct in this test.

PR #221, opened 2021-03-18
make DESC use resolution 1

LuckyMD:
Use a higher resolution for DESC to see if this improves performance.


LuckyMD:
With your change, the metrics won't run through anymore as expected, no? As then the config DESC name doesn't fit this one.

Issue #222, opened 2021-03-22
scib doesn't install with python 3.8 anymore


danielStrobl:
Due to the github actions PR and the changed dependenices, scIB does not install with Python 3.8 anymore.
scGen requires tensorflow 1.15 which is not available for python 3.8.

Do we need to have the integration tools as required dependencies?
Or would it be possible to have those as optional dependencies?

LuckyMD:
Please move to optional. We should definitely not have these as required. This might entail moving import statements into functions in `integration.py` though.

PR #223, opened 2021-03-22
Remove method dependencies

danielStrobl:
#222


LuckyMD (approved):

LuckyMD:
@mumichae when we get around to adding method tests to the GA workflow, we'll need to add an additional `pip install` for the requirements_extra.txt stuff. Just wanted to point this out here so that you're aware. For current tests this won't matter.

PR #224, opened 2021-03-22
Remove locked dependencies

danielStrobl:


requirements.txt
```diff
@@ -2,19 +2,19 @@ numpy==1.18.1
...
 rpy2==3.1.0
 anndata2ri==1.0.2
 scipy
 scikit-learn
 scikit-misc
 loompy==3.0.6
-louvain==0.6.1
-umap-learn==0.3.10
+louvain
+umap-learn
```
LuckyMD, line 16: I wonder about these 2. There is a louvain issue with >0.6.1 if I recall, and `umap-learn` should maybe be <0.5.0 if possible.

LuckyMD (approved):

LuckyMD:
Let's see if this works... GA is passing.

Issue #233, opened 2021-05-05
mapping_mouse_human.txt file


JBreunig:
Might the mapping_mouse_human.txt file be available?  I'm not sure if I'm missing it on FigShare or in one of the directories--or perhaps a script generates it?
Thanks in advance and excellent work!

LuckyMD:
Hi @JBreunig,
I guess this is the name of the gene ortholog mapping file? I don't think we uploaded that anywhere. 
@martaint could you make a PR and upload this to `scib-reproducibility` if I haven't overlooked it somewhere?

JBreunig:
Thanks! And , yes, that is the gene ortholog mapping file needed in this notebook: scib/notebooks/data_preprocessing/immune_cells/merging/Merging_all_mouse.ipynb

> In [17]:
> mapMouse_Human_orth = pd.read_csv('../mapping_mouse_human.txt', sep='\t')
> mapMouse_Human_orth.head()

LuckyMD:
The file was added in #234 and is at: `notebooks/data_preprocessing/immune_cells/merging/mapping_mouse_human.txt`

PR #234, opened 2021-05-06
add mapping orthologous genes

martaint:


lazappi:
Should this be here or `scIB-reproducibility`? I think we want to avoid duplicating stuff in general if we can.

LuckyMD:
We should put it in both... and remove this file when things are moved across to `scib-reproducibility`. This is an intermediate solution as the file was requested in #233

LuckyMD (approved):

PR #241, opened 2021-06-14
Remove redundant cc score code

mumichae:
A fresh version of PR #219

### DIFF
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -562,8 +562,11 @@ def cell_cycle(adata_pre, adata_post, batch_key, embed=None, agg_func=np.mean,
     scores_before = []
     scores_after = []
 
-    recompute_cc = recompute_cc or ('S_score' not in adata_pre.obs_keys() and 'G2M_score' not in adata_pre.obs_keys())
-    recompute_pcr = precompute_pcr_key is None or precompute_pcr_key not in adata_pre.uns_keys()
+    recompute_cc = recompute_cc \
+                   or 'S_score' not in adata_pre.obs_keys() \
+                   or 'G2M_score' not in adata_pre.obs_keys()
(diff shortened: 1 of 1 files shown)


scIB/metrics.py
```diff
@@ -558,97 +561,61 @@ def cell_cycle(adata_pre, adata_post, batch_key, embed=None, agg_func=np.mean,
...
-            int_sub = int_sub.obsm[embed] if embed is not None else int_sub.X
-        
-            if raw_sub.shape[0] != int_sub.shape[0]:
-                message = f'batch "{batch}" of batch_key "{batch_key}" '
-                message += 'has unequal number of observations before and after integration.'
-                message += f'before: {raw_sub.shape[0]} after: {int_sub.shape[0]}'
-                raise ValueError(message)
-        
+
+    recompute_cc = recompute_cc or ('S_score' not in adata_pre.obs_keys() and 'G2M_score' not in adata_pre.obs_keys())
```
LuckyMD, line 565: should this not be an `or` in there? We still can't compute scores if only one of `'S_score'` or `'G2M_score'` are there.

LuckyMD (changes requested):
one minor change. And is this tested for consistent results? Should be all the same code by the looks of things, but would be good to confirm once with precomputed stuff and once without.

scIB/metrics.py
mumichae: True, then it'd better be
```py
recompute_cc = recompute_cc or 'S_score' not in adata_pre.obs_keys() or 'G2M_score' not in adata_pre.obs_keys()
```

LuckyMD (approved):
Could you confirm that this is tested for consistency on test data before merging though?

mumichae:
I've tested it for a couple of methods and the results are identical for this branch and the master branch

PR #245, opened 2021-07-01
Moran's I as bio conservation metric

Hrovatin:
I have finally added Moran's I as bio conservation metric.

## Motivation:
Metric that captures bio preservation ** without cell type labels. Useful when no annotation is available or annotation is unreliable (e.g. for cell subtypes*).
Method:
Moran's I measures how clear spatially variable patterns genes have across embedding. Thus, if genes vary non-randomly on non-integrated data one would also want them to vary non-randomly across embedding on integrated data.
- Compute HVGs on non-integrated data.
- Compute Moran's I for these genes on integrated data. Case B: report mean(integrated_I).
- Optional: Compute Moran's I on for each non-integrated sample. Case A: report mean(integrated_I - max(sample_I))

I have used only case B so far. But if same HVGs are used across different integrations  it might be the case that B would suffice (with less computation). However, this would need to be tested.

** This corresponded strongly to ranking from other metrics on my pancreatic beta cell subtypes.
* I think this is very important point. Conservation across cell sub-types may sometimes be much worse than across cell types, as I have observed when comparing scVI and the autoencoder from scArches (not trVAE).

TODO: 

- [ ] Please check that code matches your parameter naming etc. or add any other parameters you may find necessary. For example I always recompute connectives.

- [ ] Please check that it runs in your env (Moran's I was added to Scanpy only recently) - I did not try to run it in any of your envs, just in mine

### DIFF
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -1770,6 +1770,119 @@ def graph_connectivity(adata_post, label_key):
 
     return np.mean(clust_res)
 
+def moransi_conservation(adata_pre, adata_post, batch_key, n_hvg=1000, embed='X_pca',
+                        rescale=True,compare_pre=False, hvgs=None):
+    """
+    Compute Moran's I on HVGs (defined across batches) on integrated data 
+        as a measure of bio conservation. The metric can be computed either as 
+        A.) mean difference between Morans's I on embedding of integrated data and 
+        max Moran's I of individual non-integrated batches or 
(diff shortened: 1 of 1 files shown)


Hrovatin:
Known issue: If any of the batches has <15 cells this will raise an Error since PCA with 15 components can not be computed.

mumichae:
Closed in favor of #304

PR #249, opened 2021-08-11
Allow specific pseudotime column for TI metric

mumichae:
Currently metric assumes pseudotime to be saved under 'dpt_pseudotime'. For some use-cases we would want to be able to specifiy different pseudotime columns in `.obs`.

### DIFF
--- a/scIB/metrics.py
+++ b/scIB/metrics.py
@@ -1699,7 +1699,7 @@ def kBET(adata, batch_key, label_key, embed='X_pca', type_ = None,
     return kBET_scores
 
 
-def get_root(adata_pre, adata_post, ct_key, pseudotime_key, dpt_dim=3):
+def get_root(adata_pre, adata_post, ct_key, pseudotime_key="dpt_pseudotime", dpt_dim=3):
     """
     Determine root cell for integrated adata based on unintegrated adata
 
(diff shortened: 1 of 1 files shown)


scIB/metrics.py
```diff
@@ -1698,18 +1698,31 @@ def kBET(adata, batch_key, label_key, embed='X_pca', type_ = None,
...
     return kBET_scores
 
-# determine root cell for trajectory conservation metric
-def get_root(adata_pre, adata_post, ct_key, dpt_dim=3):
-    
-    n_components, adata_post.obs['neighborhood'] = connected_components(csgraph=adata_post.obsp['connectivities'], directed=False, return_labels=True)
-    
-    start_clust = adata_pre.obs.groupby([ct_key]).mean()['dpt_pseudotime'].idxmin()
+
+def get_root(adata_pre, adata_post, ct_key, pseudotime_key, dpt_dim=3):
```
LuckyMD, line 1702: need to set a default to `'dpt_pseudotime'` here

scIB/metrics.py
```diff
@@ -1728,28 +1741,40 @@ def get_root(adata_pre, adata_post, ct_key, dpt_dim=3):
...
+    iroot, adata_post_sub2 = get_root(adata_pre_sub, adata_post_sub, label_key, pseudotime_key)
     adata_post_sub2.uns['iroot'] = iroot
     
-    sc.tl.dpt(adata_post_sub2)
-    adata_post_sub2.obs['dpt_pseudotime'][adata_post_sub2.obs['dpt_pseudotime']>1]=0
+    sc.tl.dpt(adata_post_sub2)  # stored in 'dpt_pseudotime'
+    adata_post_sub2.obs['dpt_pseudotime'][adata_post_sub2.obs['dpt_pseudotime'] > 1] = 0
     adata_post_sub.obs['dpt_pseudotime'] = 0
     adata_post_sub.obs['dpt_pseudotime'] = adata_post_sub2.obs['dpt_pseudotime']
-    adata_post_sub.obs['dpt_pseudotime'].fillna(0)
```
LuckyMD, line 1751: removing this line means you have na in the entries that were 0 before, no? That would change the metric.

LuckyMD (commented):
I would set default for get_root and check 1 line you may have removed by accident.

PR #251, opened 2021-08-19
Fix TI metric

mumichae:
Some bugs found when `pseudotime_key` is not 'dpt_pseudotime'


LuckyMD (approved):
looks good to me... not sure what the first change affects (looks like the same exact thing written differently), but fine with me

PR #257, opened 2021-09-13
fix kbet metric

mbuttner:


LuckyMD:
Is the observed rejection rate scaled in the same way that we expect? Would it be between 0 and 1 now?

mbuttner:
> Is the observed rejection rate scaled in the same way that we expect? Would it be between 0 and 1 now?

In `scIB.metrics.metrics`, we scale the kBET output with `score = 1 - np.nanmean(kBET_score)`. In `scIB`, we scaled all metrics such that 0 means poorest and 1 means best possible. An observed rejection rate (here named `kBET_score`) of 0 is best and 1 is poorest. Therefore, the scaling in the `scIB.metrics.metrics` function is correct.

LuckyMD:
Okay, cool... so the scaling would be exactly the same as for p-values then. It should probably also be somewhat correlated at least.

LuckyMD:
Let me know when this is locally tested. @mumichae can you run this with your pipeline test, or is kBET not in that environment due to the R-python compatibility question.

mbuttner:
> Okay, cool... so the scaling would be exactly the same as for p-values then. It should probably also be somewhat correlated at least.

Well, anti-correlated, I assume. Low p-values = high observed rejection rates and vice versa.

mbuttner:
> Let me know when this is locally tested. @mumichae can you run this with your pipeline test, or is kBET not in that environment due to the R-python compatibility question.

Test is currently running on pancreas data - it's running, but I have to check the results.

LuckyMD (approved):

LuckyMD:
Let's keep this on a branch for now though. @danielStrobl can you create the docker image with this branch?

mumichae:
Pipeline test just ensures that the metrics work as before. I tested all metrics as well as the metrics subset functions, they're all behaving as before

mumichae:
Should I keep the kbet code as-is for this PT and fix it in another?

mumichae:
> Let me know when this is locally tested. @mumichae can you run this with your pipeline test, or is kBET not in that environment due to the R-python compatibility question.

There wasn't much change in the API and the pipeline runs through as expected.
metrics_plot

LuckyMD:
That looks like it's changed, but still quite correlated on the test data, is that correct? Could you check for correlation?

mumichae:
I wouldn't interpret too much into this data, as it's just an implementation test without any representative batches

LuckyMD:
Yeah, I understand... still surprising as we're expecting it to essentially invert.

LuckyMD:
@mumichae could you resolve conflicts with the current master and then merge?

PR #262, opened 2021-09-16
Check dependies in functions

mumichae:
Check if optional dependencies are available in their respective functions and control behaviour when import fails.

Checks for:

- [x] all integration methods
- [x] kBET
- [x] LISI


tests/metrics/test_kbet.py
```diff
@@ -0,0 +1,15 @@
...
+def test_kbet(adata_pca):
+    score = scIB.me.kBET(
+        adata_pca,
+        batch_key='batch',
+        label_key='celltype',
+        embed='X_pca'
+    )
+    score = 1 - np.nanmean(score['kBET'])
+    LOGGER.info(f"score: {score}")
+    assert np.isnan(score)
```
LuckyMD, line 15: I think `np.nanmean` ensures this is never nan. Apparently a runtime warning is output if `np.nanmean` gets a vector of all `NaN`s.

LuckyMD (commented):
Really nice overall, but the kBET test might need some work. Have you run pipeline test on this?

tests/metrics/test_kbet.py
mbuttner: For small cell types, where kBET might not be applicable, a `NaN` value is returned. `np.mean` fails when it encounters `NaN`, `np.nanmean` ignores them.

LuckyMD: This test prob also needs to work when kBET can't be run on any celltype... This might be the case, no?

mbuttner: Indeed. In the case of all `NaN`, `np.nanmean` returns only `nan`.

LuckyMD: Ah... wait... a RunTimeWarning doesn't mean an error. All good here.

tests/metrics/test_kbet.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def test_kbet(adata_pca):
+    score = scIB.me.kBET(
+        adata_pca,
+        batch_key='batch',
+        label_key='celltype',
+        embed='X_pca'
+    )
+    score = 1 - np.nanmean(score['kBET'])
```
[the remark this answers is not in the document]
mumichae: yes, would be moved to the kbet fix

scIB/metrics/lisi.py
```diff
@@ -326,7 +326,12 @@ def lisi_graph_py(
     # create evenly split chunks if n_obs is divisible by n_chunks (doesn't really make sense on 2nd thought)
     n_splits = n_chunks - 1
     args_int = [cpp_file_path.as_posix(), mtx_file_path, dir_path, str(n_neighbors), str(n_splits), str(subset)]
-    subprocess.run(args_int)
+    try:
+        subprocess.run(args_int)
+    except Exception as e:
```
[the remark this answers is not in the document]
mumichae: I didn't know exactly which exception would be caused on a different architecture so I went with a general one for now

mumichae:
The test is tailored to the  Github actions workflow. In order for kBET to work properly, we'd need a proper R installation in the CI, which would be something for a separate workflow and PR.

LuckyMD (approved):

tests/metrics/test_kbet.py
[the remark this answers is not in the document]
LuckyMD: sounds good to me!

LuckyMD:
Thanks @mumichae !

PR #263, opened 2021-09-16
Reimplement ARI

mumichae:
Reimplement ARI to correct for overflow error in sklearn.
The sklearn implementation is not default.


scIB/metrics/ari.py
```diff
@@ -1,10 +1,12 @@
+import numpy as np
 import pandas as pd
+import scipy.special
 from sklearn.metrics.cluster import adjusted_rand_score
 
 from scIB.utils import checkAdata, checkBatch
 
 
-def ari(adata, group1, group2):
+def ari(adata, group1, group2, implementation='sklearn'):
```
LuckyMD, line 9: why do we default to an implementation that potentially will not work? I would probably do the opposite here.

scIB/metrics/ari.py
```diff
@@ -14,20 +16,31 @@ def ari(adata, group1, group2):
...
-    elif isinstance(group1, pd.Series):
-        group1 = group1.tolist()
-
-    if isinstance(group2, str):
-        checkBatch(group2, adata.obs)
-        group2 = adata.obs[group2].tolist()
-    elif isinstance(group2, pd.Series):
-        group2 = group2.tolist()
+    group1 = adata.obs[group1].to_numpy()
+    group2 = adata.obs[group2].to_numpy()
```
LuckyMD, line 23: What happened to the `checkBatch` calls? Also is this no longer able to take a `pd.Series`? The expected input type should probably be in the documentation above.

LuckyMD (commented):
Just some questions on what the default should be and input checks.

scIB/metrics/ari.py
mumichae: `checkBatch` is called no matter the input type and I wouldn't support pd.Series, as we never use it and it wouldn't make much sense if we also provide an anndata as input. For pd.Series input, we should have a different function signature.

LuckyMD: makes sense! With the documentation it's all good :).

LuckyMD (approved):

LuckyMD:
This looks really good, thanks so much @mumichae !

PR #264, opened 2021-09-18
Check kbet lisi

mbuttner:


scIB/metrics/kbet.py
```diff
@@ -78,7 +78,7 @@ def kBET(
         embed='X_pca',
         type_=None,
         hvg=False,
-        subsample=0.5,
+        subsample=0.5, #non-functional
```
mumichae, line 81: I'd suggest we remove that option completely, i.e also from `kBET_single` and `metrics`

mumichae (commented):
Looking good so far. I think we could be more radical in removing never used parameters and comments. Are changes for the LISI functions also planned?

mbuttner:
> Looking good so far. I think we could be more radical in removing never used parameters and comments. Are changes for the LISI functions also planned?

Thanks, I checked the LISI functions and did not find things that I would change.

mumichae:
OK, then I would cleanup the code, which shouldn't change the score. That'll be subject to a new PR.

mumichae (approved):

mbuttner:
Awesome, thank you!

PR #265, opened 2021-09-19
Refactor metrics

mumichae:
Clean up LISI and kBET metrics.

- split iLISI and cLISI
- kBET scaling inside function
- Move kBET NeighborsError to kBET function


LuckyMD:
Did we really want to split cLISI and iLISI? These are computed in the same function, no?

mbuttner:
> Did we really want to split cLISI and iLISI? These are computed in the same function, no?

I wouldn't split cLISI and iLISI, either. In fact, I would rather merge the computation of both cLISI and iLISI, because the steps are the same, just a different label to compute the metric on - it would save quite some computation time.

mumichae:
From what I can see, the only time saved is during recomputing the kNN graph. If we move the control of preprocessing outside of the `ilisi_graph` and `clisi_graph` functions, that time could be halved. Also, I didn't think that was the most time consuming step (but would have to double-check). The actual LISI computation (using the C code) is done separately for each metric anyway, so splitting up the code shouldn't change much.

Additionally, the old `lisi_graph` function works as before, computing both scores, so we could use that for the `metrics` function if we wanted. I just thought it would be best to have the same level of control over `iLISI` an `cLISI` as we would have over all the other metrics when using the API.

mbuttner:
> From what I can see, the only time saved is during recomputing the kNN graph. If we move the control of preprocessing outside of the `ilisi_graph` and `clisi_graph` functions, that time could be halved. Also, I didn't think that was the most time consuming step (but would have to double-check). In the original implementation all the LISI computations (using the C code) are called separately for `iLISI` and `cLISI` anyway, so splitting up the code shouldn't change much.
> 
> Additionally, the old `lisi_graph` function works as before, computing both scores, so we could use that for the `metrics` function if we wanted. I just thought it would be best to have the same level of control over `iLISI` an `cLISI` as we would have over all the other metrics when using the API.

The thing with the two LISI functions is that the `batch_key`/`label_key` info is only used at the very end of `compute_simpson_index_graph`:
```
batch = batch_labels[knn_idx]
B = convertToOneHot(batch, n_batches)
sumP = np.matmul(P, B)  # sum P per batch
simpson[i[0]] = np.dot(sumP, sumP)  # sum squares
```
So the most efficient way to compute both `iLISI` and `cLISI` would be to add both batch and label keys at this piece of code. However, then both functions are invariably intertwined.

scIB/metrics/lisi.py
```diff
@@ -181,62 +200,121 @@ def lisi_graph(
...
 
-    clisi_score = lisi_graph_py(adata=adata_tmp, batch_key=label_key,
-                                n_neighbors=k0, perplexity=None, subsample=subsample,
-                                multiprocessing=multiprocessing, nodes=nodes, verbose=verbose)
+    return ilisi
+
+
+def clisi_graph(
+        adata,
+        batch_key,
```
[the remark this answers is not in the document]
mbuttner: Ah OK, indeed!

scIB/metrics/kbet.py
```diff
@@ -121,77 +115,91 @@ def kBET(
...
+                    score = kBET_single(
+                        matrix=matrix,
+                        batch=adata_sub.obs[batch_key],
+                        knn=nn_index_tmp + 1,  # nn_index in python is 0-based and 1-based in R
+                        verbose=verbose,
+                        k0=k0
+                    )
+                except NeighborsError:
+                    print('Not enough neighbours')
+                    score = 0
```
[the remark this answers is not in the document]
mumichae: Thanks for noticing

tests/metrics/test_ilisi.py
```diff
@@ -0,0 +1,36 @@
+from tests.common import *
+
+
+def test_ilisi_full(adata):
```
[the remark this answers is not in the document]
mumichae: Yes, the CI runs through without error

mumichae:
> The thing with the two LISI functions is that the `batch_key`/`label_key` info is only used at the very end of `compute_simpson_index_graph`:
> 
> ```
> batch = batch_labels[knn_idx]
> B = convertToOneHot(batch, n_batches)
> sumP = np.matmul(P, B)  # sum P per batch
> simpson[i[0]] = np.dot(sumP, sumP)  # sum squares
> ```
> 
> So the most efficient way to compute both `iLISI` and `cLISI` would be to add both batch and label keys at this piece of code. However, then both functions are invariably intertwined.

So that means that the old implementation of the score wasn't the most efficient either, because we call the `lisi_graph_py` and with that the `compute_simpson_index_graph` twice anyway. 

Would it be better if we extracted the `compute_simpson_index_graph` functionality outside of `lisi_graph_py`, compute it once and then pass the output for cLISI and iLISI computations?
In that case I'd suggest we optimise the `lisi_graph` function, but still keep separate cLISI and iLISI functions for the API. The latter could have an option to preprocess the simpson index and pass it to those functions. I hope that isn't overkill...

mumichae:
I think this PR should be ready to merge and we can address LISI computation optimisations in a different PR. Any comments or wishes to the current state?

gokceneraslan:
This PR also fixes, #266, so it'd be really great to see it merged 😄

mumichae:
@mbuttner Could you go over the latest changes and approve if the PR is ready to be merged?

scIB/metrics/metrics.py
```diff
@@ -149,6 +148,14 @@ def metrics_all(
...
+        isolated_labels_f1_=True,
+        trajectory_=True,
+        nmi_=True,
+        ari_=True,
+        cell_cycle_=True,
+        kBET_=True,
+        ilisi_=True,
+        clisi_=True,
         **kwargs
     )
```
LuckyMD, line 160: Could add a deprecation comment for this as well... we don't use it anymore as it only works for python methods.

scIB/metrics/metrics.py
```diff
@@ -177,9 +184,10 @@ def metrics(
         n_isolated=None,
         graph_conn_=False,
         kBET_=False,
-        kBET_sub=0.5,
+        subsample=0.5,
```
LuckyMD, line 187: I thought this parameter was removed from the kBET method?

scIB/metrics/metrics.py
```diff
@@ -199,14 +207,13 @@ def metrics(
...
     if nmi_ or ari_:
-        print('Clustering...')
         res_max, nmi_max, nmi_all = opt_louvain(
             adata_int,
             label_key=label_key,
             cluster_key=cluster_key,
             function=nmi,
             plot=False,
-            verbose=verbose,
+            verbose=False,
```
LuckyMD, line 216: why remove verbosity passing here?

scIB/metrics/metrics.py
```diff
@@ -325,58 +332,50 @@ def metrics(
...
-    # results['iLISI'] = ilisi_score
-    # results['cLISI'] = clisi_score
-
     if lisi_graph_:
-        print('LISI graph score...')
-        ilisi_raw, clisi_raw = lisi_graph(
+        clisi_ = True,
+        ilisi_ = True
+
+    if clisi_:
```
LuckyMD, line 351: any reason to not just keep this in `lisi_graph_`? I guess this also allows separate clisi and ilisi runs?

LuckyMD (commented):
I love the lisi tests! Just 1 or 2 questions... otherwise this looks good from my side. If you answer the questions, we can merge.

scIB/metrics/metrics.py
LuckyMD: this is only passed to LISI now?

scIB/metrics/metrics.py
mumichae: Deprecation for the LISI parameters?

scIB/metrics/metrics.py
mumichae: yes, that's why I renamed it. We never used it for kBET.

scIB/metrics/metrics.py
mumichae: The output wasn't very informative (it's only preprocessing), so I removed it

scIB/metrics/metrics.py
mumichae: Yes, the user then also has the option to run one and not the other. I just added that for consistency with the rest of the behaviour. The old `lisi_graph_` parameter should just work as before

LuckyMD (approved):

scIB/metrics/metrics.py
LuckyMD: Maybe it's sth for a future "debug" option for verbosity. Let's leave for now.

LuckyMD:
Once @mbuttner reviews, you can merge. All good from my side now.

mbuttner (approved):
Looks good to me.

PR #268, opened 2021-09-27
Check metrics

mumichae:
Making sure all metrics are doing what we expect them to.

### DIFF
--- a/scIB/metrics/graph_connectivity.py
+++ b/scIB/metrics/graph_connectivity.py
@@ -3,25 +3,29 @@ import pandas as pd
 from scipy.sparse.csgraph import connected_components
 
 
-def graph_connectivity(adata_post, label_key):
+def graph_connectivity(adata, label_key):
     """"
-    Metric that quantifies how connected the subgraph corresponding to each batch cluster is.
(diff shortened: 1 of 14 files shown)


scIB/metrics/graph_connectivity.py
```diff
@@ -26,6 +26,6 @@ def graph_connectivity(adata, label_key):
             connection='strong'
         )
         tab = pd.value_counts(labels)
-        clust_res.append(tab[0] / sum(tab))
+        clust_res.append(tab.max() / sum(tab))
```
LuckyMD, line 29: should this be argmax or max?

LuckyMD (changes requested):
Looks good overall!

scIB/metrics/graph_connectivity.py
mumichae: max, if we want the number of nodes in the largest connected component, argmax if we want its label

mumichae:
This PR shouldn't have changed any metrics except for graph connectivity. Here are the scores from the pipeline on test data
image

LuckyMD (approved):

PR #269, opened 2021-09-27
Update highly_variable_genes.py

danielStrobl:
Clean up hvg_overlap metric, remove commented out code, remove unused pre-storage of highly variable genes


scIB/metrics/highly_variable_genes.py
```diff
@@ -28,32 +28,20 @@ def hvg_overlap(adata_pre, adata_post, batch, n_hvg=500, verbose=False):
...
-            #    print(adata_post_list[i].obs[batch][0]+' has less than the specified number of genes')
-            #    print('Number of genes: '+str(adata_post_list[i].n_vars))
-            # hvg_pre = sc.pp.highly_variable_genes(adata_pre_list[i], flavor='cell_ranger', n_top_genes=n_hvg_tmp, inplace=False)
-            tmp_pre = hvg_pre_list[batch_var]  # adata_pre_list[i].var.index[hvg_pre['highly_variable']]
-            hvg_post = sc.pp.highly_variable_genes(adata_post_list[i], flavor='cell_ranger', n_top_genes=n_hvg_tmp,
-                                                   inplace=False)
-            tmp_post = adata_post_list[i].var.index[hvg_post['highly_variable']]
-            n_hvg_real = np.minimum(len(tmp_pre), len(tmp_post))
-            overlap.append((len(set(tmp_pre).intersection(set(tmp_post)))) / n_hvg_real)
+    hvg_pre_list = precompute_hvg_batch(adata_pre, batch, hvg_post)
```
mumichae, line 31: What's the difference between the hvg function here vs the function from preprocessing?

scIB/metrics/highly_variable_genes.py
```diff
@@ -28,32 +28,20 @@ def hvg_overlap(adata_pre, adata_post, batch, n_hvg=500, verbose=False):
...
+
+    for i in range(len(adata_post_list)):  # range(len(adata_pre_list)):
+        sc.pp.filter_genes(adata_post_list[i], min_cells=1)  # remove genes unexpressed (otherwise hvg might break)
+        batch_var = adata_post_list[i].obs[batch][0]
+        n_hvg_tmp = len(hvg_pre_list[batch_var])
```
mumichae, line 36: Make sure that you are indexing correctly. In Pandas, 0 is usually assumed to be a label, not a numeric location. Also, what is batch_var supposed to represent?

I think it would be better to iterate over objects rather than indices

```suggestion
    for adata in adata_post_list:
        sc.pp.filter_genes(adata, min_cells=1)  # remove genes unexpressed (otherwise hvg might break)
        batch = adata.obs[batch].unique()[0]
        n_hvg_tmp = len(hvg_pre_list[batch])
```

Or even better, something like this

```
for batch, adata, hvgs in zip(...):
    ... 
```

mumichae (changes requested):
Would be great to have some more documnetation on the HVG computation and HVG overlap score. Could you add some docstrings?

About the code, it might be better to make it more readable, so anyone can understand what is being done. I added some suggestions.

scIB/metrics/highly_variable_genes.py
danielStrobl: this one returns the hvgs per batch and not genes that are variable in all batches

scIB/metrics/highly_variable_genes.py
danielStrobl: `batch_var` is just the name of the batch, it doesn't really matter which element of this column is chosen as they are all the same. I think for pandas Series the 0-th element should be fine, no?

scIB/metrics/highly_variable_genes.py
mumichae: OK, is `hvg_pre_list` a suitable name or should we rename to to `hvgs_by_batch`

scIB/metrics/highly_variable_genes.py
mumichae: 0 should be fine, as long as the index doesn't have a different index.

I'd still recommend simplifying the names if it's not too much trouble.

scIB/metrics/highly_variable_genes.py
LuckyMD: i think this is fine... it's `hvgs_by_batch_pre` otherwise... a bit long...

LuckyMD (approved):
Looks good to me. Some naming could be improved, but fine.

mumichae:
This PR has been lying around for a while. @danielStrobl Could you please check if the code is correct? Then we should be able to merge

danielStrobl:
looks good to me

mumichae (approved):

PR #272, opened 2021-10-03
Packaging

mumichae:
First proper scib package for pypi.


scib/integration.py
```diff
@@ -1,24 +1,24 @@
-#!/bin/env python
-
 ### D. C. Strobl, M. Müller; 2019-07-23
```
LuckyMD, line 1: might be worth adding a few more names to this list?

mumichae: This is just the integration module. We have the package author list in the `setup.cfg`

setup.cfg
```diff
@@ -1,3 +1,78 @@
...
+author_email = malte.luecken@helmholtz-muenchen.de, michaela.mueller@helmholtz-muenchen.de
+description = Evaluate single cell integration methods
+long_description = file: README.md
+long_description_content_type = text/markdown
+license = MIT
+url = https://github.com/theislab/scib
+project_urls =
+    Pipeline = https://github.com/theislab/scib-pipeline
+    Reproducibility = https://theislab.github.io/scib-reproducibility
+Bug Tracker = https://github.com/theislab/scib/issues
```
LuckyMD: do you need an underline here between "Bug" and "Tracker"?

setup.cfg
```diff
@@ -1,3 +1,78 @@
...
+keywords = benchmark, single cell, data integration
```
LuckyMD: maybe "benchmarking" instead of "benchmark"

setup.cfg
```diff
@@ -1,3 +1,78 @@
...
+classifiers =
+    Development Status :: 3 - Alpha
+    Intended Audience :: Developers
```
LuckyMD, line 21: Also people who are using data integration methods: So "Developers", "Data analysts"

setup.cfg
```diff
@@ -1,3 +1,78 @@
...
+    numba
+    scanpy>=1.5
+    anndata>=0.7.2
+    h5py<3
+    rpy2>=3
+    anndata2ri
+    scipy
+    scikit-learn
+    scikit-misc
+    loompy==3.0.6
```
LuckyMD, line 50: why do we need loompy? Is this a core dependency? Or is this related to conos or sth?

setup.cfg
```diff
@@ -1,3 +1,78 @@
...
+    llvmlite
+
+[options.packages.find]
+where = scib
+include = metrics
+
+#[options.package_data]
+#scib =
+#    resources/*.txt
+#    knn_graph/*
```
LuckyMD: Why not include package data here?

README.md
```diff
@@ -24,14 +24,24 @@ batches of gene expression and chromatin accessibility data.
...
-We created the python package called `scIB` that uses `scanpy` to streamline the integration of single-cell datasets and
-evaluate the results. The evaluation of integration quality is based on a number of metrics.
+We created the python package called `scib` that uses `scanpy` to streamline the integration of single-cell datasets and
+evaluate the results. For evaluating the integration quality it provides a number of metrics.
+
+### Requirements
+
++ Linux or UNIX system
++ Python >= 3.7
++ R >= 3.6
```
LuckyMD, line 36: also R<4.0?

scib/integration.py
```diff
@@ -297,14 +295,14 @@ def runScanvi(adata, batch, labels):
     return adata
 
 
-def runMNN(adata, batch, hvg=None):
+def mnnpy(adata, batch, hvg=None):
```
LuckyMD, line 298: Must call this mnn I think... mnnpy is the name of the module, but not the integration method.

scib/metrics/trajectory.py
```diff
@@ -60,44 +62,55 @@ def get_root(
...
-
+    adata_post_ti2.uns['iroot'] = iroot
+
+    sc.tl.dpt(adata_post_ti2)  # stored in 'dpt_pseudotime'
+    adata_post_ti2.obs.loc[adata_post_ti2.obs['dpt_pseudotime'] > 1, 'dpt_pseudotime'] = 0
+    adata_post_ti.obs['dpt_pseudotime'] = 0
+    adata_post_ti.obs['dpt_pseudotime'] = adata_post_ti2.obs['dpt_pseudotime']
+    adata_post_ti.obs['dpt_pseudotime'].fillna(0, inplace=True)
+
+    adata_post_ti.obs['batch'] = adata_pre_ti.obs['batch']
```
LuckyMD, line 101: Is this okay, @danielStrobl? Why should this not be stored in `adata_post_ti` already?

setup.cfg
```diff
@@ -1,18 +1,21 @@
 [metadata]
 name = scib
 version = 0.2.0
-author = Malte Luecken, Maren Buettner, Daniel Strobl, Michaela Mueller
-author_email = malte.luecken@helmholtz-muenchen.de, michaela.mueller@helmholtz-muenchen.de
-description = Evaluate single cell integration methods
+description = Benchmark tools for single cell data integration
 long_description = file: README.md
 long_description_content_type = text/markdown
+author = Malte Luecken, Maren Buettner, Daniel Strobl, Michaela Mueller
```
LuckyMD, line 7: please add middle initials to match paper format

setup.cfg
```diff
@@ -1,18 +1,21 @@
...
 license = MIT
 url = https://github.com/theislab/scib
 project_urls =
     Pipeline = https://github.com/theislab/scib-pipeline
     Reproducibility = https://theislab.github.io/scib-reproducibility
-Bug Tracker = https://github.com/theislab/scib/issues
-keywords = benchmark, single cell, data integration
+    Bug Tracker = https://github.com/theislab/scib/issues
+keywords =
+    benchmark
```
LuckyMD, line 16: benchmarking

LuckyMD (commented):
Overall this looks very nice. Just two main points:
1. Do we have a parallel PR in `scib-pipeline` to ensure all of the scripts are using the correct function names? Or are we running our own module in a deprecated manner by default?
2. @danielStrobl can we assign batch in the trajectory score as you did? Why should this not already be in the second object?

setup.cfg
mumichae: It should be fine without. See here for a demo https://test.pypi.org/project/scib/

setup.cfg
mumichae: Yes, I guess it was method specific so I'll remove it

README.md
LuckyMD: What do you think about pining to <4.0 for now? Maybe some versions of R methods will no longer work? I'm not sure...

setup.cfg
```diff
@@ -1,24 +1,25 @@
...
 name = scib
 version = 0.2.0
-description = Benchmark tools for single cell data integration
+description = Evaluating single-cell data integration methods
 long_description = file: README.md
 long_description_content_type = text/markdown
-author = Malte Luecken, Maren Buettner, Daniel Strobl, Michaela Mueller
-author_email = malte.luecken@helmholtz-muenchen.de, michaela.mueller@helmholtz-muenchen.de
+author = Malte D Luecken, Maren Buettner, Daniel C Strobl, Michaela F Mueller
+author_email = malte.luecken@helmholtz-muenchen.de
```
LuckyMD, line 8: you can keep your helmholtz email on there if you want. And if you amend... add dots to the initials please ^^

setup.cfg
```diff
@@ -47,7 +48,6 @@ install_requires =
     scipy
     scikit-learn
     scikit-misc
-    loompy==3.0.6
```
LuckyMD: Did you test if removing this is okay?

LuckyMD (approved):

scib/metrics/trajectory.py
mumichae: I'd suggest we assume that the batches should match before the function is called. Ideally this should be something that the integration method wrapper or the user take care of.
Alternatively I'd throw an error message if batches mismatch. We might risk incorrectly overwriting the batch information otherwise.

setup.cfg
mumichae: I recreated the scib-pipeline environment from scratch and everything ran through without issue. It must have been some integration method specific dependency, as it also doesn't affect the github actions test.

PR #273, opened 2021-10-04
add per batch trajectory score

danielStrobl:
Add functionality to compute trajectory score per batch and take the mean and not compute the score overall.

### DIFF
--- a/scIB/metrics/trajectory.py
+++ b/scIB/metrics/trajectory.py
@@ -4,6 +4,7 @@ import scanpy as sc
 import pandas as pd
 
 from .utils import RootCellError
+from ..utils import checkBatch
 
 
 def get_root(
@@ -69,7 +70,7 @@ def trajectory_conservation(
         adata_post,
         label_key,
         pseudotime_key="dpt_pseudotime",
-	batch_key=None
+        batch_key=None
 ):
     """
     :param adata_pre: unintegrated adata
@@ -99,16 +100,17 @@ def trajectory_conservation(
 
     adata_post_ti.obs['batch'] = adata_pre_ti.obs['batch']
 
-    
     if batch_key == None:
         pseudotime_before = adata_pre_ti.obs[pseudotime_key]
         pseudotime_after = adata_post_ti.obs['dpt_pseudotime']
         correlation = pseudotime_before.corr(pseudotime_after, 'spearman')
         return (correlation + 1) / 2  # scaled
     else:
+        checkBatch(adata_pre, batch_key)
+        checkBatch(adata_post, batch_key)
         corr = pd.Series()
         for i in adata_pre_ti.obs[batch_key].unique():
-            pseudotime_before = adata_pre_ti.obs[adata_pre_ti.obs[batch_key]==i][pseudotime_key]
-            pseudotime_after = adata_post_ti.obs[adata_post_ti.obs[batch_key]==i]['dpt_pseudotime']
+            pseudotime_before = adata_pre_ti.obs[adata_pre_ti.obs[batch_key] == i][pseudotime_key]
+            pseudotime_after = adata_post_ti.obs[adata_post_ti.obs[batch_key] == i]['dpt_pseudotime']
             corr[i] = pseudotime_before.corr(pseudotime_after, 'spearman')
-        return (corr.mean() + 1) / 2 # scaled
+        return (corr.mean() + 1) / 2  # scaled
--- a/scIB/metrics/trajectory.py
+++ b/scIB/metrics/trajectory.py
@@ -68,6 +68,7 @@ def trajectory_conservation(
         adata_post,
         label_key,
         pseudotime_key="dpt_pseudotime"
+	batch_key=False
 ):
     """
     :param adata_pre: unintegrated adata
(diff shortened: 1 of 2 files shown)


scIB/metrics/trajectory.py
```diff
@@ -67,37 +68,61 @@ def trajectory_conservation(
...
-    adata_post_sub.obs['dpt_pseudotime'] = adata_post_sub2.obs['dpt_pseudotime']
-    adata_post_sub.obs['dpt_pseudotime'].fillna(0, inplace=True)
-
-    pseudotime_before = adata_post_sub.obs['dpt_pseudotime']
-    pseudotime_after = adata_pre_sub.obs[pseudotime_key]
-    correlation = pseudotime_before.corr(pseudotime_after, 'spearman')
-
-    return (correlation + 1) / 2  # scaled
-
+    if batch_key==False:
```
mumichae, line 86: Use `batch_key == None` instead of `False`

scIB/metrics/trajectory.py
```diff
@@ -67,37 +68,61 @@ def trajectory_conservation(
...
+            sc.tl.dpt(adata_post_sub2_tmp)  # stored in 'dpt_pseudotime'
+            adata_post_sub2_tmp.obs.loc[adata_post_sub2_tmp.obs['dpt_pseudotime'] > 1, 'dpt_pseudotime'] = 0
+            adata_post_sub_tmp.obs['dpt_pseudotime'] = 0
+            adata_post_sub_tmp.obs['dpt_pseudotime'] = adata_post_sub2_tmp.obs['dpt_pseudotime']
+            adata_post_sub_tmp.obs['dpt_pseudotime'].fillna(0, inplace=True)
+            
+            pseudotime_before = adata_post_sub_tmp.obs['dpt_pseudotime']
+            pseudotime_after = adata_pre_sub_tmp.obs[pseudotime_key]
+            corr[i] = pseudotime_before.corr(pseudotime_after, 'spearman')
+        return (corr.mean() + 1) / 2 # scaled
```
mumichae, line 128: I'm assuming this part of the code is a duplicate of the code above, just called per batch. It would be great if you could put this into a separate function and call that function twice instead. That way it should be easier to read

scIB/metrics/trajectory.py
```diff
@@ -67,37 +68,61 @@ def trajectory_conservation(
...
     :param adata_post: integrated adata
     :param label_key: column in `adata_pre.obs` of the groups used to precompute the trajectory
     :param pseudotime_key: column in `adata_pre.obs` in which the pseudotime is saved in.
         Column can contain empty entries, the dataset will be subset to the cells with scores.
+    :param batch_key: set to batch key if if you want to compute the trajectory metric by batch
     """
     # subset to cells for which pseudotime has been computed
     cell_subset = adata_pre.obs.index[adata_pre.obs[pseudotime_key].notnull()]
     adata_pre_sub = adata_pre[cell_subset]
     adata_post_sub = adata_post[cell_subset]
```
mumichae, line 85: Could we have better naming e.g. `adata_pre_ti`? `_sub` is a bit general...

scIB/metrics/trajectory.py
```diff
@@ -67,37 +68,61 @@ def trajectory_conservation(
...
-
-    pseudotime_before = adata_post_sub.obs['dpt_pseudotime']
-    pseudotime_after = adata_pre_sub.obs[pseudotime_key]
-    correlation = pseudotime_before.corr(pseudotime_after, 'spearman')
-
-    return (correlation + 1) / 2  # scaled
-
+    if batch_key==False:
+        try:
+            iroot, adata_post_sub2 = get_root(adata_pre_sub, adata_post_sub, label_key, pseudotime_key)
```
mumichae, line 88: Could we rename to something more meaningful e.g. `adata_pre_root`?

mumichae (changes requested):
Looking good, but I'd suggest avoiding duplicated code as much as possible and tweaking the variable names to avoid any potential bugs.

LuckyMD:
@danielStrobl This version should not compute the trajectory per batch. It should only do the Spearman correlation per batch at the end. The trajectory should still be correct across all batches. That's why you don't need that much duplicated code... only at the very end you need to subset to do spearman correlation per batch and take the mean.

scIB/metrics/trajectory.py
```diff
@@ -60,44 +61,51 @@ def get_root(
...
+    pseudotime_before = adata_post_ti.obs['dpt_pseudotime']
+    pseudotime_after = adata_pre_ti.obs[pseudotime_key]
     correlation = pseudotime_before.corr(pseudotime_after, 'spearman')
-
-    return (correlation + 1) / 2  # scaled
-
+    if batch_key == None:
+        return (correlation + 1) / 2  # scaled
+    else:
+        corr = pd.Series()
```
LuckyMD, line 106: This part needs a check if the `batch_key` actually exists

scIB/metrics/trajectory.py
```diff
@@ -60,44 +61,51 @@ def get_root(
...
+        for i in batch_key:
```
LuckyMD, line 107: According to the documentation `batch_key` is a string that is the name of an `adata.obs` column. Here, it looks like it is `adata.obs[batch_key].unique()`. This needs to be adapted.

LuckyMD (changes requested):
How you deal with `batch_key` needs to be adapted. Also, please test the code on pancreas and compare to without batch key.

scIB/metrics/trajectory.py
danielStrobl: yeah, should have actually tested the code

LuckyMD:
Great! Have you got the results comparison for a few methods so we can assesss the impact of using the batch structure?

LuckyMD:
@danielStrobl, nice! That means most importantly the results are consistent especially for the top performing methods :). Ready to merge now @mumichae !

PR #274, opened 2021-10-18
Add **kwds to be able to access method-specific arguments

gokceneraslan:
I needed to access arguments like `pynndescent_random_state` in bbknn, so I thought it might make sense to add **kwds to these methods so that users can specify method-specific arguments.


LuckyMD:
Thanks @gokceneraslan! This makes a lot of sense... @mumichae could you merge this with #272 ?

PR #279, opened 2021-12-02
Documentation

mumichae:
Update docstrings and create sphinx docs for readthedocs


scib/metrics/clustering.py
```diff
@@ -6,27 +6,40 @@
...
-    returns:
-        res_max: resolution of maximum score
-        score_max: maximum score
-        score_all: `pd.DataFrame` containing all scores at resolutions. Can be used to plot the score profile.
-        clustering: only if `inplace=False`, return cluster assignment as `pd.Series`
-        plot: if `plot=True` plot the score profile over resolution
+    Louvain clustering optimised against a metric over multiple resolutions.
+
+    :params adata: anndata object
+    :params label_key: name of column in adata.obs containing biological labels to be
```
LuckyMD, line 26: params still there

scib/metrics/ari.py
```diff
@@ -7,15 +7,16 @@
...
+    Adjusted Rand Index.
     The function is symmetric, so group1 and group2 can be switched
-    For single cell integration evaluation the scenario is:
-        predicted cluster assignments vs. ground-truth (e.g. cell type) assignments
+    For single cell integration evaluation the comparison is between predicted cluster
+    assignments and the ground truth (e.g. cell type)
 
     :param adata: anndata object
     :param group1: string of column in adata.obs containing labels
     :param group2: string of column in adata.obs containing labels
```
lazappi, line 18: I know this is symmetric so it doesn't matter but I would give these clearer names (and make them consistent across functions).

scib/metrics/clustering.py
```diff
@@ -6,27 +6,40 @@
...
+    Louvain clustering optimised against a metric over multiple resolutions.
+
+    :param adata: anndata object
+    :param label_key: name of column in adata.obs containing biological labels to be
+        optimised against
+    :param cluster_key: name of column to be added to adata.obs during clustering.
+        Will be overwritten if exists and ``force=True``
+    :param function: function that computes the cost to be optimised over. Must take as
+        arguments ``(adata, group1, group2, **kwargs)`` and returns a number for maximising
+    :param resolutions: list if resolutions to be optimised over. If ``resolutions=None``,
```
lazappi, line 32: Typo `if` -> `of`

scib/metrics/clustering.py
```diff
@@ -6,27 +6,40 @@
 from .nmi import nmi
 
 
-def opt_louvain(adata, label_key, cluster_key, function=None, resolutions=None,
-                use_rep=None,
-                inplace=True, plot=False, force=True, verbose=True, **kwargs):
+def opt_louvain(
```
lazappi, line 9: Should this be in the preprocessing module rather than metrics?

scib/metrics/lisi.py
```diff
@@ -32,13 +33,12 @@ def lisi(
...
-        covariate_key: variable to compute iLISI on
-        cluster_key: variable to compute cLISI on
-    return:
-        pd.DataFrame with median cLISI and median iLISI scores (following the harmony paper)
+    Compute iLISI and cLISI scores
+
+    :param matrix: matrix from adata to calculate on
+    :param covariate_key: variable to compute iLISI on
+    :param cluster_key: variable to compute cLISI on
+    :return: Tuple of  median iLISI and median cLISI scores
```
lazappi, line 41: Typo: extra space

scib/metrics/lisi.py
```diff
@@ -663,15 +661,19 @@ def Hbeta(D_row, beta):
 
 def convertToOneHot(vector, num_classes=None):
```
lazappi, line 662: This function uses camelCase but I think snake_case is used for most things

scib/metrics/lisi.py
```diff
@@ -692,8 +694,7 @@ def convertToOneHot(vector, num_classes=None):
     return result.astype(int)
 
 
-# DEPRECATED
-# This code scales clisi incorrectly!
+@deprecated
```
lazappi, line 697: Is there a reason to keep this or can it be deleted?

scib/metrics/metrics.py
```diff
@@ -147,36 +187,72 @@ def metrics(
         adata_int,
         batch_key,
         label_key,
-        hvg_score_=False,
+        embed='X_pca',
         cluster_key='cluster',
         cluster_nmi=None,
         ari_=False,
         nmi_=False,
```
lazappi, line 194: Why do these arguments have underscores?

scib/metrics/metrics.py
```diff
@@ -147,36 +187,72 @@ def metrics(
...
+    :param isolated_labels_: whether to compute both isolated label scores
+    :param isolated_labels_f1_: wehther to compute isolated label score based on F1 score of clusters vs labels
+    :param isolated_labels_asw_: whether to compute isolated label score based on ASW
+    :param n_isolated: max number of batches per label for label to be considered as isolated
+    :param graph_conn_: whether to compute graph connectivity score
+    :param trajectory_: whether to compute trajectory score
+    :param kBET_: whether to compute kBET score
+    :param lisi_graph_: whether to compute both cLISI and iLISI
+    :param clisi_: whether to compute cLISI
+    :param ilisi_: whether to compute iLISI
```
lazappi, line 253: Could these link to the individual metric functions?

scib/metrics/silhouette.py
```diff
@@ -44,20 +45,21 @@ def silhouette_batch(
...
     :param metric: see sklearn silhouette score
     :param scale: if True, scale between 0 and 1
     :param return_all: if True, return all silhouette scores and label means
         default False: return average width silhouette (ASW)
     :param verbose:
     :return:
-        average width silhouette ASW
-        mean silhouette per group in pd.DataFrame
-        Absolute silhouette scores per group label
+        average silhouette width ASW  (always)
```
lazappi, line 60: Typo: extra space

scib/preprocessing.py
```diff
@@ -264,18 +264,26 @@ def scale_batch(adata, batch):
...
+        adata,
+        batch,
+        target_genes=2000,
+        flavor='cell_ranger',
+        n_bins=20,
+        adataOut=False,
+        n_stop=8000,
+        min_genes=500,
+        step_size=1000
+):
```
lazappi, line 277: I think the scanpy function now does batch HVG selection. Could you use that instead?

lazappi (changes requested):
I think this is heading in the right direction. Some general comments:

- In the rendered version the workflow image seems to be missing
- Should the installation instructions include PyPI? I think there is only from GitHub at the moment.
- I think the table of contents and organisation could do with a bit of work. I like the format the scanpy docs have where there is an overview page for each module and then individual pages for functions.
- I think some of the links to the manuscript are still for the preprint. Probably worth double checking all of these.
- Some functions still need documenting, mostly in the integration and preprocessing modules
- I definitely think it's worth making sure argument names are standardised across functions that do the same thing (i.e. all the metrics functions). This makes it much easier for users when you only need to learn what they mean once.
- Not sure how much work this would be but it would be great to add examples to more functions. These can be really useful for showing exactly what needs to be provided to arguments.
- I found a few small typos and there are probably some more. Not a priority at this stage but worth checking once you are pretty happy with things. Usually it's easier to do on the rendered version than the code.

lazappi (changes requested):
I think this is heading in the right direction. Some general comments:

- In the rendered version the workflow image seems to be missing
- Should the installation instructions include PyPI? I think there is only from GitHub at the moment.
- I think the table of contents and organisation could do with a bit of work. I like the format the scanpy docs have where there is an overview page for each module and then individual pages for functions.
- I think some of the links to the manuscript are still for the preprint. Probably worth double checking all of these.
- Some functions still need documenting, mostly in the integration and preprocessing modules
- I definitely think it's worth making sure argument names are standardised across functions that do the same thing (i.e. all the metrics functions). This makes it much easier for users when you only need to learn what they mean once.
- Not sure how much work this would be but it would be great to add examples to more functions. These can be really useful for showing exactly what needs to be provided to arguments.
- I found a few small typos and there are probably some more. Not a priority at this stage but worth checking once you are pretty happy with things. Usually it's easier to do on the rendered version than the code.

scib/preprocessing.py
mumichae: I think we discussed that at some point but ultimately decided on our own implementation. @LuckyMD would this be something to reconsider?

LuckyMD: When we decided to implement this, the scanpy function didn't yet exist... then goekcen wrote it before we were done. It should be more or less the same. For consistency I think we should keep this. Also because I have other code in a reproducibility github somewhere that uses it.

scib/metrics/clustering.py
LuckyMD: It's part of a few metrics, so I think prob not. Preprocessing is for data preprocessing rather than metric helper functions.

scib/metrics/ari.py
LuckyMD: agreed

LuckyMD: I think this should be cluster and label for ARI, and not "batch".

scib/metrics/clustering.py
mumichae: We tried to move it to preprocessing, however we get circular inputs, because the function relies on the metrics package, but is at the same time needed for some metrics.

scib/metrics/lisi.py
mumichae: This is only used in the `scib.me.lisi.lisi` function, but we use `scib.me.lisi.lisi_graph` where scaling is replaced with newer code. Can we also consider `scib.me.lisi.lisi` as deprecated, since we aren't using it anywhere (I marked it as such)?

scib/metrics/metrics.py
mumichae: I think it was initially to make sure that they don't get confused with the functions themselves, but I'm not sure if that's actually a problem. Might consider changing these parameters, which would also require changes in scib-pipeline

scib/metrics/ari.py
mumichae: I'll keep this for now and deal with it in another PR

scib/preprocessing.py
mumichae: Keeping this for now

mumichae:
Thanks @lazappi for the in-depth feedback! I tried to address most points, but will keep standardising the label names and adding examples for a new PR.

LuckyMD:
Hey! This is looking really nice overall. I went through the readthedocs now and have quite a few comments, which can mostly be quite easily addressed though I think. The main thing is the organization of the table of contents with two "package:scib" bullet points. We just need to organize that a bit better. Everything else is a minor point on a function docstring.

* [ ]Table of contents overview with 2 “package:scib:
    * [ ] probably good to remove the “Metrics” part from the first package:scib bullet point and then call the whole thing “Overview” or sth like that
* [ ] Resources should also include the “scib” package
* [ ] HVG intersect: add doctoring “legacy approach to HVG selection only using HVG intersections between all batches”* * [ ] Scran normalisation -> add ref?
* [ ] read_conos -> needs a doctoring
* [ ] Can we link to Scanpy documentation for `Scanpy.tl.score_genes_cell_cycle` and to scran documentation or CRAN repo for the `scran` ref?
* [ ] Is it possible to have separate pages for the Package:scib submodules?
* [ ] DESC, SAUCIE functions docstring needs a link to the original method (not just parameterisation taken from here link)
* [ ] Can we add package versions to the integration methods? For scanVI the scvi-tools link is to a newer version of the package that is not compatible with our code I believe.
* [ ] scvi url is to a fork of scvi-tools. Is this really the one we’re using?
* [ ] Add to metrics overview that more info can be found in the paper (and link to it).
* [ ] “Edit on Github” button not working?
* [ ] Not sure scib.metrics.silhouette explanation works. 1 indicates distinct, compact clusters; 0 indicates overlapping clusters, -1 indicates core-periphery (non-cluster) structure
* [ ] cLISI/iLISI graph metrics should have some more info: “Local Inverse Simpson’s Index metrics adapted from url to run on all types of integration outputs via shortest path-based distance computation on single-cell knn graphs.”
* [ ] return value for cLISI_graph should be more specific: “median over what?”
* [ ] cLISI “params adata” what is that?
* [ ] graph connectivity docstring is wrong. It should say “corresponding to each cell type label is”. Also, can we put some newlines here to distinguish “final score”from “calculate per label”?
* [ ] batch ASW: nice explanation, but let’s try to use maths rather than code here (can take this from paper)
* [ ] PCR comparison: “R2Var value of PCR” is not clear. Can you just write “variance contribution”?
* [ ] kBET: need to say this is computed per cell type and then averaged already in the description (not just in return value)
* [ ] Can we add default values to the documentation? => mainly important for k0 in LISI scores
* [ ] wow… metrics function docstrings are awesome!
* [ ] for metrics call subfunctions, could you mention where to find info on the kwargs that are passed on?

LuckyMD:
> Afaik, default values are already included in the autodoc. Or do you want more specific explanations on which values to use?

I didn't see any in the readthedocs link...

mumichae:
> > Afaik, default values are already included in the autodoc. Or do you want more specific explanations on which values to use?
> 
> I didn't see any in the readthedocs link...

Screenshot from 2022-04-22 15-04-43

Default values are shown in the function signature.

LuckyMD:
Ah, I was looking at the parameter explanation. Okay,all good. Thanks!

ivirshup (approved):
Looks fantastic! A few minor points, but I think it would be good to merge.

---------------

A few general minor things which would be nice, but could/ should be separate PRs:

* Type information + [autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) would be nice
* It would be great if you can installation instructions for setting up an isolated environment (probably conda), especially due to the mixed language dependencies
  * Kinda related: wouldn't you need Seurat for `read_seurat`?
* I would strongly recommend turning on PR builds for docs: https://docs.readthedocs.io/en/stable/pull-requests.html

scib/metrics/__init__.py
```diff
@@ -1,16 +1,32 @@
...
+| + NMI cluster/label                                | + kBET                                               |
+|                                                    |                                                      |
+| + Highly variable gene overlap                     |                                                      |
+|                                                    |                                                      |
+| + Isolated label ASW                               |                                                      |
+|                                                    |                                                      |
+| + Isolated label F1                                |                                                      |
+|                                                    |                                                      |
+| + Trajectory conservation                          |                                                      |
++----------------------------------------------------+------------------------------------------------------+
```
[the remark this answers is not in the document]
mumichae: Good point. I wanted to provide an overview of the different metrics, as shown in the README, but I guess this table would be best for the README itself.

docs/source/scib_metrics.rst
```diff
@@ -0,0 +1,89 @@
...
+
+
+The remaining functions call the :func:`scib.metrics.metrics` for
+
+Furthermore, :func:`scib.metrics.metrics()` is wrapped by convenience functions with preconfigured subsets of metrics
+based on expected computation time:
+
++ :func:`scib.me.metrics_fast()` only computes metrics that require little preprocessing
++ :func:`scib.me.metrics_slim()` includes all functions of :func:`scib.me.metrics_fast()` and adds clustering-based metrics
++ :func:`scib.me.metrics_all()` includes all metrics
```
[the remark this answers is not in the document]
mumichae: Thanks, must have missed those. I wasn't sure if it would be better to use the full function name or not, but I guess it's a bit redundant, as more information will be shown once the user clicks on the reference.

scib/metrics/__init__.py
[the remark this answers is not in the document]
mumichae: I moved this table to the README now.

PR #294, opened 2022-03-07
Update dependency versions

mumichae:
Remove version pins for numpy and h5py.


mumichae:
For some reason Github actions is returning NMI scores slightly larger than 1 during testing. I couldn't reproduce this error offline.

```
    def test_nmi_trivial(adata):
        score = scib.me.nmi(adata, 'celltype', 'celltype')
>       assert score == 1
E       assert 1.0000000000000002 == 1
E         +1.0000000000000002
E         -1
```

Zethson:
Great, thank you!

PR #295, opened 2022-03-08
Remove `,` in `bool` var

johnarevalo:
`clisi_` should not be a `tuple`, but a `bool` value.

### DIFF
--- a/scib/metrics/metrics.py
+++ b/scib/metrics/metrics.py
@@ -328,7 +328,7 @@ def metrics(
         kbet_score = np.nan
 
     if lisi_graph_:
-        clisi_ = True,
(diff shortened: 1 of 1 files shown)


mumichae:
Thanks for spotting this! Does it break any code?

johnarevalo:
Not that I am aware of. I'm suggesting it just for completeness. It helps to avoid confusions as in this snippet

```python
clisi_ = True,
print('Enabled' if clisi_ else 'Disabled')
Enabled

clisi_ = False,
>>> print('Enabled' if clisi_ else 'Disabled')
Enabled
```

mumichae (approved):

PR #297, opened 2022-04-01
Addressing installation issues

mumichae:
Addressing issues listed under https://github.com/theislab/scib-pipeline/issues/23

> Things to improve installation experience:
> * [x] pin rpy2 3.4.2
>   * install specifically via conda (if possible)
> * [x] scib: try-except around ro.r('library(Seurat)’) to report that library installs were problematic

### DIFF
--- a/scib/exceptions.py
+++ b/scib/exceptions.py
@@ -5,3 +5,9 @@ class IntegrationMethodNotFound(ModuleNotFoundError):
                        "please make sure you install it manually."
         super().__init__(self.message)
 
+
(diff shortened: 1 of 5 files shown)


mumichae:
@LuckyMD I guess including exception handling for all other R packages in rpy2 would be helpful?

LuckyMD:
Yes, I think so

LuckyMD (approved):

LuckyMD:
We prob also need this for the `scib-pipeline` integration scripts, no?

PR #302, opened 2022-04-29
Fix LISI warning error

mumichae:
The warning message if LISI encounters a case with too few neighbours throws an error.

### DIFF
--- a/scib/metrics/lisi.py
+++ b/scib/metrics/lisi.py
@@ -430,43 +430,42 @@ def compute_simpson_index_graph(
     simpson = np.zeros(len(chunk_ids))
 
     # loop over all cells in chunk
-    for i in enumerate(chunk_ids):
+    for i, chunk_id in enumerate(chunk_ids):
         # get neighbors and distances
         # read line i from indices matrix
-        get_col = indices[i[1]]
+        get_col = indices[chunk_id]
 
         if get_col.isnull().sum() > 0:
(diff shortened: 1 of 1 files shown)


scib/metrics/lisi.py
```diff
@@ -430,43 +430,42 @@ def compute_simpson_index_graph(
     simpson = np.zeros(len(chunk_ids))
 
     # loop over all cells in chunk
-    for i in enumerate(chunk_ids):
+    for i, chunk_id in enumerate(chunk_ids):
```
mumichae, line 433: Changed notation for better readability

scib/metrics/lisi.py
```diff
@@ -430,43 +430,42 @@ def compute_simpson_index_graph(
...
         # get neighbors and distances
         # read line i from indices matrix
-        get_col = indices[i[1]]
+        get_col = indices[chunk_id]
 
         if get_col.isnull().sum() > 0:
             # not enough neighbors
-            print(i[1] + " has not enough neighbors.")
-            simpson[i[0]] = 1  # np.nan #set nan for testing
+            print(f'Chunk {chunk_id} does not have enough neighbors. Skipping...')
```
mumichae, line 440: Bugfix here. F-strings used to avoid type mismatch

mumichae:
Thanks for the quick review!

PR #303, opened 2022-05-02
Add metric morans i

MxMstrmn:


scib/metrics/metrics.py
```diff
@@ -493,25 +485,36 @@ def metrics(
...
     else:
         trajectory_score = np.nan
+    
+    if moransi_: 
+        print("Moran's I score...")
+        moransi_score = morans_i(
+            adata, 
+            adata_int, 
+            batch_key=batch_key
+        )
```
LuckyMD, line 495: looks like your auto-formatting didn't strike here... inconsistency with comma after last parameter or not? ^^

scib/metrics/morans_i.py
```diff
@@ -0,0 +1,149 @@
...
+import pandas as pd
+
+from scib.preprocessing import hvg_batch
+
+
+def morans_i(
+    adata_pre,
+    adata_post,
+    batch_key,
+    n_hvg=1000,
```
LuckyMD, line 12: 1000 hvgs sounds like a lot for this as default. Is there any reason for this? Can it be reduced?

scib/metrics/morans_i.py
```diff
@@ -0,0 +1,149 @@
...
+        computed across batches).
+    :return: The resulting metric.
+    """
+    # Prepare adatas
+    # Get integrated connectivities for Moran's I
+    adata_post = adata_post.copy()
+    sc.pp.neighbors(adata_post, use_rep=embed)
+    # Prepare pre data
+    adata_pre = adata_pre.copy()
+    adata_pre.obs[batch_key] = adata_pre.obs[batch_key].astype("category")
```
LuckyMD, line 44: Please add a call to `check_sanity()` or the `check_adata` or `check_batch` util functions from here: https://github.com/theislab/scib/blob/f35b0c990f435e1c4bfab4a6fb38451712d010f6/scib/utils.py#L25

scib/metrics/morans_i.py
```diff
@@ -0,0 +1,149 @@
...
+        Helper function for computing Moran's I on HVGS that are non-constant
+        and adding them to var
+        :param data: adata, result is added to var
+        :param hvgs: Genes for which to compute Moran's I
+        """
+        # Make sure that genes used for computation are non-constant
+        cond = (
+            np.array(np.mean(data[:, hvgs].X, axis=0))
+            != np.array(data[0, hvgs].X.todense())
+        ).squeeze()
```
LuckyMD, line 65: mean equal to first value could happen if the genes are not constant, no? Why not define this via standard deviation?

scib/metrics/morans_i.py
```diff
@@ -0,0 +1,149 @@
...
+            # Compute HVG for each sample
+            sc.pp.highly_variable_genes(
+                adata_sample,
+                flavor="cell_ranger",
+                batch_key="study_sample",
+                n_top_genes=2000,
+            )
+            adata_sample_scl = adata_sample.copy()
+            # PP each sample: scaling, pca, neighbours
+            sc.pp.scale(adata_sample_scl, max_value=10)
```
LuckyMD, line 97: just out of curiosity: any reason for scaling here specifically or out of habit?

scib/metrics/morans_i.py
```diff
@@ -0,0 +1,149 @@
...
+        # Table of Moran's I-s across batches
+        batch_mis = []
+        for sample, data in adatas_sample.items():
+            mi = data.var["morans_i"]
+            mi.name = sample
+            batch_mis.append(mi)
+        batch_mis = pd.concat(batch_mis, axis=1)
+        # Difference with integrated data
+        mi_diffs = adata_post.var["morans_i"] - batch_mis.max(axis=1)
+        avg_mi_diff = mi_diffs.mean()
```
LuckyMD, line 132: I don't entirely understand this part. You are expecting optimal integration to mean that Moran's I is the same same in the full integration as the max in any individual batch? Is the max used due to the possibility of having unique cell types in a single batch? Is this affected by cell type composition differences across batches?

scib/metrics/morans_i.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # Rescale so that it is between [0,1] where 1 is better
+        # Moran's I will be in [-1,1] and thus difference can be in [-2,2]
+        if rescale:
+            res = (avg_mi_diff + 2) / 4
```
LuckyMD, line 137: Another note on rescaling. Here it seems that a difference of 2 (perfect auto-correlation of markers in the integrated data but max per-batch auto-correlation is -1) is optimal. I'm not sure this should be the case. Wouldn't the idea be to detect the same spatial correlation structure in a single batch as across all of them? I guess you want a `1 - (avg_mi_diff + 2) / 4` here, no? 

Also, should a Moran's I of -1 and 0 be distinguished for the comparison at all? Both mean no biologically meaningful information is encoded, no?

scib/metrics/morans_i.py
```diff
@@ -0,0 +1,149 @@
...
+        else:
+            res = avg_mi_diff
+
+    else:
+        # Moran's I on integrated data
+        compute_mi(adata_post, hvgs)
+        res = adata_post.var["morans_i"].mean()
+        if rescale:
+            # Moran's I will be in [-1,1]
+            res = (res + 1) / 2
```
LuckyMD, line 147: also here: should a Moran's I of 0 be better than one of -1? Could that difference only be due to sparsity of the marker?

LuckyMD (commented):
I still have some questions about the rationale for the rescaling and comparison pre and post. It seems in the end, you default to not comparing the score to pre-integration? Do we have a ground truth that 1 is desirable for all HVGs that we're using? This seems a bit tough for me to argue for. Why not go for the pre-post comparison instead? Also, I'm not sure why we should care about distinguishing values of -1 and 0.

scib/metrics/metrics.py
MxMstrmn: This is corrected on #304

scib/metrics/morans_i.py
MxMstrmn: I agree, this can be reduced - maybe to 100?

scib/metrics/morans_i.py
MxMstrmn: You are right! std is better!

scib/metrics/morans_i.py
MxMstrmn: Most of the code was adapted from Karin. @Hrovatin, do you have an opinion here?

scib/metrics/morans_i.py
MxMstrmn: -1 means perfectly dispersed, if we are looking for a clustered signal we could do `max(0, score)` to have the score in `[0,1]` from the beginning - what do you think? 

Wrt to the `1 - ... `, I totally agree. Minimal distances should yield a high score.

scib/metrics/morans_i.py
MxMstrmn: I guess the max gives an upper value on the degree at which an individual gene can be clustered. I changed it in #304 such that only worse morans I scores are counted - if you agree that the max is an upper bound per gene, then a difference > 0 would mean that the integration was successful in maintaining this trait of the data. 

To compare different integrations, I believe that that `compare_pre=False` computation would be sensible as scores are directly comparable (their mean).

scib/metrics/morans_i.py
MxMstrmn: EDIT: I changed it such as to only consider "bad" differences, please check in #304

scib/metrics/morans_i.py
MxMstrmn: I changed it to scores >0 as we are mostly interested in clusterings.

PR #304, opened 2022-05-02
Feat: add moransi

MxMstrmn:
Updated version of moran's I PR with pre-commits 

A new approach to integrate #245 

- [x] Transfer code to new branch (based on main)
- [x] Which preprocessing needed?
- [x] Pytest function
- [x] Use scanpy implementation (https://github.com/scverse/scanpy/blob/master/scanpy/metrics/_morans_i.py,. https://scanpy.readthedocs.io/en/latest/generated/scanpy.metrics.morans_i.html#scanpy-metrics-morans-i)
- [ ] Moran’s I → Moran’s I comparison
- [x] Parameters of function
- [ ] Integrate in metrics wrapper methods


MxMstrmn:
Hi @michalk8, do you have an idea why GA fail here? The file locally has no syntax errors and also passes the pre-commits I added as part of this PR. Do not know how to handle this best / fix it.

michalk8:
Done. Also, I'd seriously consider refactoring the CI to be more up-to-date.
Interestingly, AST checker pre-commit did not catch this.

MxMstrmn:
Why does `f"{variable=}"` fail the parse, is that not normal python syntax?

MxMstrmn:
I did not set the CI up, I guess @mumichae and I would be very glad for your input improve this :)

michalk8:
> Why does f"{variable=}" fail the parse, is that not normal python syntax?

It is, just from Python 3.8 (would deprecate 3.7 since numpy support ended last December), hence the refactoring.

MxMstrmn:
Thanks so much, @michalk8!

mumichae:
> Done. Also, I'd seriously consider refactoring the CI to be more up-to-date.
> Interestingly, AST checker pre-commit did not catch this.

Are you referring to the .github/workflow content to be updated based on a newer configuration template? I wasn't aware there was an update, what consequences would the current code have of not updated?

Would be happy to update, however that would be in a separate PR.

michalk8:
> Are you referring to the .github/workflow content to be updated based on a newer configuration template? I wasn't aware there was an update, what consequences would the current code have of not updated?

Would do the following:
- remove 3.7 CI, add 3.9 (unless you need to support it)
- use pre-commits (add e.g. black) instead of running raw flake8 (this PR was)
- add tests coverage
- not sure how much work, but macOS test CI would be nice to have

The first 2 points I'd say are the one to focus on.

mumichae:
Thanks a lot for the suggestions @michalk8 !

I think we should definitely make things more consistent, if we start using pre-commit (which we hadn't before). But I think this should be separated from Moran's I. I'll open a new issue with a description of what we'd need for pre-commit integration and would suggest we get that merged before the new metric.

MxMstrmn:
Updated version of #303

mumichae (approved):

PR #306, opened 2022-05-03
Fix duplicate cols scanorama

MxMstrmn:
Close https://github.com/theislab/scib-pipeline/issues/32

Most important is: 0b397b3f8fa365ac87a0a06c08d94121ce7be794
Other two are pre-commits + formatting which is missing in the current scib/ scib-pipeline packages.


MxMstrmn:
@michalk8, do you mind checking how I remove the cols here 0b397b3f8fa365ac87a0a06c08d94121ce7be794 - could be that you know a more elegant way :) 

@LuckyMD, @mumichae, this code fixes the error reported [here](https://github.com/theislab/scib-pipeline/issues/32) but might obscure what is going wrong under the hood. When I debugged this issue on my data the adata files in `corrected` contained 3 columns called `mean` and `std`, respectively. Maybe this helps you already what happens here. In any case, I included a print statement to make the deletion transparent.

michalk8:
> @michalk8, do you mind checking how I remove the cols here https://github.com/theislab/scib/commit/0b397b3f8fa365ac87a0a06c08d94121ce7be794 - could be that you know a more elegant way :)

This will also keep the first duplicated columns (from left to right):
```python
for attr in ("obs", "var"):
    df = getattr(adata, attr)
    dup_mask = df.columns.duplicated()
    if dup_mask.any():
        print(f"Deleting duplicated keys `{list(df.columns[dup_mask].unique())}` from `adata.{attr}`.")
        setattr(adata, attr, df.loc[:, ~dup_mask])
```

MxMstrmn:
Love it @michalk8, thanks a lot! :)

mumichae (changes requested):
Thanks for catching this. Is it a common thing to have duplicate columns?

I think we may have tried to address this issue with the `scib.utils.merge_adatas()` function. It would be good to move this fix to or update `scib.utils.merge_adatas()` and use that function.

MxMstrmn:
This happens inside the scib-pipeline so I cannot really answer if this
happens a lot - do you think moving the fix would still solve this issue
where it occured? 

On May 12, 2022, Michaela Mueller ***@***.***> wrote:
> @mumichae requested changes on this pull request.
>
> Thanks for catching this. Is it a common thing to have duplicate
> columns?
>
> I think we may have tried to address this issue with the
> scib.utils.merge_adatas() function. It would be good to move this fix
> to or update scib.utils.merge_adatas() and use that function.
>

—
Reply to this email directly, view it on GitHub
<https://github.com/theislab/scib/pull/306#pullrequestreview-970736876>,
or unsubscribe <https://github.com/notifications/unsubscribe-
auth/AF6VGRN5MQFMM3SKD72B3F3VJTYI5ANCNFSM5VAFRR5A>.
You are receiving this because you authored the thread.Message ID:
***@***.***>

mumichae:
What exactly do duplicate columns look like? Are they columns with exactly the same name? I'm curious how that happens in pandas, as I'm trying to reproduce the initial error.

mumichae:
I moved the duplicate column removal to the `scib.utils.merge_adata` function and call it in `scanorama` as well as in the `scib.preprocessing.scale_batch` functions.

@MxMstrmn Could you test if this code is working for your data? If you have a way to generate duplicate columns, I'd love to add a test to pytest

mumichae (approved):

MxMstrmn:
Yes, I will check it and add a test!

PR #307, opened 2022-05-04
Add pre-commit

### LINKED ISSUE #305: Use pre-commit
### Start using pre-commit

- [x] include working pre-commit config file
- [x] update syntax of all python files
- [x] include dependencies to 'dev' section of setup.cfg
- [x] include README section mentioning the installatio of scib with dev dependencies for developers

### Suggestions from @michalk8 :

> - [x] remove 3.7 CI, add 3.9 (unless you need to support it)
> - [x] use pre-commits (add e.g. black) instead of running raw flake8 (this PR was)
> - [x] add tests coverage
> - [x] not sure how much work, but macOS test CI would be nice to have

with emphasis on the first 2 points

### Notes:

+ This PR targets pre-commit integration and syntax updates. Separate points in the list above should be addressed in separate PRs.
+ We still need to support python 3.7, as this is the version used in scib-pipeline (expect users working with it already)

MxMstrmn:
I added a pre-commit config which worked well for me in the past. I added the two specifier from the CI to flake8:
```yaml
args: 
        - --select=E9,F63,F7,F82
        - --max-line-length=127
```
No sure if these should be kept in the future - what do you think, @michalk8? 

Also, I am not sure how to adjust the `setup.cfg` some help would be cool here :) 

Close #305

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Build and Test
+name: Test
 
 on:
     push:
@@ -28,34 +28,6 @@ jobs:
             run: |
                 pre-commit run --color=always --show-diff-on-failure --all-files
 
-    build:
-        runs-on: ${{ matrix.os }}
-        strategy:
-            matrix:
-                python: [3.7, 3.9]
-                os: [ubuntu-latest, macos-latest]
-
-        steps:
-        -   uses: actions/checkout@v2
-            with:
-                fetch-depth: 0
-        -   name: Set up Python ${{ matrix.python }}
-            uses: actions/setup-python@v2
-            with:
-                python-version: ${{ matrix.python }}
-        -   name: Install pip dependencies
-            run: |
-                python -m pip install --upgrade pip
-                pip install build twine
-
-        -   name: Build a binary wheel and a source tarball
-            run: |
-                python -m build --sdist --wheel --outdir dist/
-
-        -   name: Twine check
-            run: |
-                twine check dist/*
-
     test:
         runs-on: ${{ matrix.os }}
         strategy:
@@ -68,8 +40,20 @@ jobs:
             uses: actions/setup-python@v2
             with:
                 python-version: ${{ matrix.python }}
-        -   name: Install package
+
+        -   name: Install build dependencies
             run: |
+                python -m pip install --upgrade pip
+                pip install build twine
+
+        -   name: Build a binary wheel and a source tarball
+            run: |
+                python -m build --sdist --wheel --outdir dist/
+                twine check dist/*
+
+        -   name: Install package with test dependencies
+            run: |
+                pip install dist/*.whl
                 pip install '.[test]'
 
         -   name: Test with pytest


michalk8:
> No sure if these should be kept in the future - what do you think, @michalk8?

Ideally, this would be specified in `setup.cfg`, along with other config values (e.g. for black/isort).
Furthermore, the pre-commit should be enabled in the CI as well instead of doing this: https://flake8.pycqa.org/en/latest/user/configuration.html
Minor pet-peeve: the `.yml` file name is slightly deceptive (seems related to packaging, but what it does is lint/test), would rename to `test.yml` and change the name to, e.g., `name: Test`.

mumichae:
You mean separate github actions workflows? Or jobs?

mumichae:
> > No sure if these should be kept in the future - what do you think, @michalk8?
> 
> Ideally, this would be specified in `setup.cfg`, along with other config values (e.g. for black/isort). Furthermore, the pre-commit should be enabled in the CI as well instead of doing this: https://flake8.pycqa.org/en/latest/user/configuration.html Minor pet-peeve: the `.yml` file name is slightly deceptive (seems related to packaging, but what it does is lint/test), would rename to `test.yml` and change the name to, e.g., `name: Test`.

Thanks, I addressed the latter part.

Are you suggesting to have a separate flake8 configuration in `setup.cfg` or is `pre-commit` enough?

michalk8:
> Are you suggesting to have a separate flake8 configuration in setup.cfg or is pre-commit enough?

I'd add the specific arguments you call in the CI to the `[flake8]` in `setup.cfg` and not pass them in pre-commit (only the flake8 hook should be there).

mumichae:
OK, I'll do that. What difference does that make?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,47 @@
...
+    -   id: isort
+        additional_dependencies: [toml]
+        args: [--order-by-type]
+-   repo: https://github.com/myint/autoflake
+    rev: v1.4
+    hooks:
+    -   id: autoflake
+        exclude: .*notebooks.*
+        args: [--in-place, --remove-all-unused-imports, --remove-unused-variable, --ignore-init-module-imports]
+-   repo: https://gitlab.com/pycqa/flake8
```
mumichae, line 26: Is this hook necessary? Especially after transferring the into to the setup.cfg?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,47 @@
...
+    -   id: flake8
+        args: 
+        - --select=E9,F63,F7,F82
+        - --max-line-length=127
+        exclude: .*notebooks.*
+        additional_dependencies: [flake8-tidy-imports, flake8-docstrings, flake8-rst-docstrings, flake8-comprehensions, flake8-bugbear, flake8-logging-format, flake8-blind-except, flake8-builtins, flake8-pytest-style, flake8-mock, flake8-string-format]
+-   repo: https://github.com/mwouts/jupytext
+    rev: v1.13.6
+    hooks:
+    -   id: jupytext
```
mumichae, line 38: Do we need jupytext? The package should not have any notebooks

.pre-commit-config.yaml
```diff
@@ -0,0 +1,47 @@
...
+        args: [--in-place, --remove-all-unused-imports, --remove-unused-variable, --ignore-init-module-imports]
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.4
+    hooks:
+    -   id: flake8
+        args: 
+        - --select=E9,F63,F7,F82
+        - --max-line-length=127
+        exclude: .*notebooks.*
+        additional_dependencies: [flake8-tidy-imports, flake8-docstrings, flake8-rst-docstrings, flake8-comprehensions, flake8-bugbear, flake8-logging-format, flake8-blind-except, flake8-builtins, flake8-pytest-style, flake8-mock, flake8-string-format]
```
mumichae, line 34: Do we need all the dependencies?

mumichae (commented):
A couple of questions on whether certain hooks are needed.

.pre-commit-config.yaml
michalk8: Afaik `additional_dependencies` needs to still be specified in the pre-commit config, since they are plugins.
If you wanna prune them, take a look at https://github.com/scverse/cookiecutter-scverse/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.pre-commit-config.yaml#L34-L40

.pre-commit-config.yaml
MxMstrmn: No this can be deleted

mumichae:
OK, with help of @michalk8, we added the precommit config from the scverse cookiecutter  and cleaned up the workflows to have a deploy and test with linting. Reviews are welcome

Zethson (approved):
So I prefer also having an explicit workflow that only builds the package (without publishing it). Nice to catch errors when suddenly something went wrong with the dependencies or whatever.

Busy AF and only glanced for 20 seconds, but what I saw LGTM

PR #312, opened 2022-05-15
Add documentation for metrics preprocessing

mumichae:
https://scib.readthedocs.io/en/doc-metrics_input/


LuckyMD:
A couple of questions for the image overview for metrics inputs:
1 .How are the upper and lower lists of metrics different?
2. Isn't clustering performed by the ARI and NMI metrics? Why is it a necessary input?

Overall the diagram isn't super clear or particularly explained in the text.

mumichae:
1. Metrics in the upper half are run because on the metrics output directly in contrast to metrics in the lower half that require preprocessing. E.g. ASW should be be used with PCA for feature output or with the integrated embedding for embedding output.   
2. Clustering is not part of the ARI or NMI functions, so it is a necessary preprocessing step

I'll address some more explanation of the figure in the text.

ivirshup (approved):
- [ ] Remove `pip install .` from installation, since that's more for dev docs which are below

- [ ]  Installation docs could get a table of contents

- [ ]  It would be nice if the legend on the metrics preprocessing figure was on the right, close to where those colors are used

A broader suggestion for reorganization:

- [ ]  Move the section "preprocessing data for metrics" to a "user guide" section

- [ ]  Move all the "functions" section from all modules to one unified API page

This would all moving text around, I don't think anything would need to be rewritten.

PR #315, opened 2022-05-20
Set revision for kBET installation

johnarevalo:
Latest commit in kBET package breaks backward compatibility for R=3.6.1.

https://github.com/theislab/kBET/blob/2e2b09179bba4808903d04e4425f9d5cdbeaa989/R/kBET.R#L195

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -34,7 +34,7 @@ Additionally, in order to run the R package ``kBET``, you need to install it thr
 
 .. code-block:: R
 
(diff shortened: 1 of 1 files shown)


mumichae:
Hi @johnarevalo ,
I tried to reproduce the issue with R 3.6.1 using the latest version of kBET, but couldn't. Could you provide the error message that you are getting?

johnarevalo:
Thanks for your reply @mumichae , I did a clean installation and the message dissapeared.

PR #317, opened 2022-05-22
add harmony

adamgayoso:


LuckyMD:
@mumichae If we add this, how would you propose to resolve this in scib-pipeline? Maybe calling this harmony_py?

scib/integration.py
```diff
@@ -10,6 +10,29 @@
...
 
 
+def harmony(adata, batch, hvg=None, **kwargs):
+    """Harmony wrapper function
+
+    Based on `harmony-pytorch <https://github.com/lilab-bcb/harmony-pytorch>`_ version 0.1.7
+
+    :param adata: preprocessed ``anndata`` object
+    :param batch: batch key in ``adata.obs``
+    :param hvg: list of highly variables to subset to. If ``None``, the full dataset will be used
```
mumichae, line 20: Parameter not used, but this seems to be a general for other methods as well, so I'd address this in a separate PR

mumichae (approved):
Thanks for integrating this and sorry for the late review. One minor comment on the hvg paramater that is not used, bug that's a general issue to be adressed somewhere else

PR #325, opened 2022-07-07
Plotting functions

mumichae:
Provide python function for plotting metrics


LuckyMD:
Oooh, you are building a python version of Marta's plotting functionality? Or is this a different visualization? Do you have an example?

mumichae:
It's a simplified version of Marta's plot. I thought that some basic plotting functions should be part of the package, but was starting out with something simple

metrics_plot_overall

Overall score is optional:

metrics_plot

LuckyMD:
Very nice! Would be good to check if the overall score calculation is the same as for marta's code :)

mumichae:
The overall score is just a simple mean of all metrics and equally weighted. I haven't yet addressed more sophisticated weighting, but I think that would be good to address as a separate problem.

LuckyMD:
Okay... I tested different weighting approaches and converged on min-max scaling in the paper... I wouldn't test too many for now, but should be consistent with what we have for the R function. Can keep mean as well as that's useful for when you have only few methods tested.

Issue #324, opened 2022-07-02
SCANVI bug


adamgayoso:
https://github.com/theislab/scib/blob/ad40b62d4c8b3a5712ef2bc46853588a2b8cf149/scib/integration.py#L281

My apologies -- this should be `unlabeled_category` and not `unknown_category`

I could try to get a PR next week but you all might be faster

PR #327, opened 2022-07-09
Fix scanvi bug

mumichae:
Addresses #324 and https://github.com/theislab/scib-pipeline/issues/41

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,7 +54,7 @@ jobs:
         -   name: Install package with test dependencies
             run: |
                 pip install dist/*.whl
-                pip install '.[test,scanorama,scvi]'
+                pip install '.[test]'
 
         -   name: Test with pytest
             run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,4 +85,9 @@ jobs:
 
         -   name: Test with pytest
             run: |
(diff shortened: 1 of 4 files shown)


LuckyMD (approved):
This looks good... but to be clear: you added scvi and scanvi tests but don't run them in the test workflow now? I guess standard installation is without integration methods, so that's probably fine... is this tested anywhere or only manually now?

mumichae:
I do test the integration methods, but in a separate step, so it runs in parallel and also updates the code coverage

LuckyMD:
okay, cool! Didn't see the separate test in this PR, but I guess it's there from before :).

.github/workflows/test.yml
```diff
@@ -54,14 +54,38 @@ jobs:
...
+                python-version: ${{ matrix.python }}
+
+        -   name: Install dependencies
+            run: |
+                python -m pip install --upgrade pip
+                pip install '.[test,scanorama,scvi]'
+
+        -   name: Test with pytest
+            run: |
+                python -m pytest --cov --cov-append --cov-report=term-missing -k integration -vv
```
mumichae, line 88: Separate GA integration test ;)

LuckyMD: ah, yes... for some reason i only read this as code coverage ^^. but the `integration` parameter in there is key ;)

PR #334, opened 2022-09-06
Bump rp2y dependency

### LINKED ISSUE #322: Bump rpy2 version?
Would it be possible/how much work would it be to bump the **rpy2** dependency version?

The reason I ask is that **rpy2** `v3.4.2` is only compatible with R 3.6 which was released in 2019 and is three versions old. This means that any conda environment which includes **scib** is limited to an old version of R and R packages (for Bioconductor at least). It would be great if this restriction was removed so that more recent versions of R could be used with **scib**.

mumichae:
Unpin rpy2 dependency and make sure test run through without failing. This will close #322 

Skipping pytest on macos (due to unresolved segmentation fault)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,11 +57,11 @@ jobs:
                 pip install '.[test]'
 
         -   name: Test with pytest
-            if: ${{ matrix.os }} != 'macos-latest'
+            if: matrix.os != 'macos-latest'
             run: |
                 python -m pytest --cov --cov-append --cov-report=term-missing --ignore=tests/integration/ -vv
         -   name: Upload coverage
-            if: ${{ matrix.os }} != 'macos-latest'
+            if: matrix.os != 'macos-latest'
             env:
                 CODECOV_NAME: ${{ matrix.os }}-${{ matrix.python }}
             run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
         strategy:
             matrix:
                 python: [3.7, 3.9]
-                os: [ubuntu-latest, macos-latest]
+                os: [ubuntu-latest]
         steps:
         -   uses: actions/checkout@v2
         -   name: Set up Python ${{ matrix.python }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
         strategy:
             matrix:
                 python: [3.7, 3.9]
-                os: [ubuntu-latest]
+                os: [ubuntu-latest, macos-latest]
         steps:
         -   uses: actions/checkout@v2
         -   name: Set up Python ${{ matrix.python }}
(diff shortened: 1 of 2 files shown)


Zethson:
I don't see where the PR unpins rpy2?

Zethson:
> That was done on main (which is bad Ik) and then the CI broke. I was just cleaning up the aftermath that could have been related to the version bump. Should I perhaps rename the PR?

Ahhh, no worries. I usually recommend to add a reason for why you're pinning something (igraph in this case) IF you're not pinning everything (would be the alternative). Can later help figure out why igraph<0.10 was required in the first place. If this pin is not necessary consider getting rid of it :)

mumichae:
igraph 0.10 was only released today and the louvain functions was throwing errors, so I pinned it. However, I see that leidenalg also restricts to igraph<0.10, so in that case igraph no longer needs to be pinned

setup.cfg
```diff
@@ -58,9 +59,10 @@ install_requires =
 	scikit-learn
 	scikit-misc
 	louvain
+    leidenalg
 	umap-learn
 	pydot
-	python-igraph
+	igraph
```
LuckyMD, line 65: is `igraph` now the official install? I can't find the old `python-igraph` as separate anymore, so i guess so?

.github/workflows/test.yml
```diff
@@ -57,9 +57,11 @@ jobs:
                 pip install '.[test]'
 
         -   name: Test with pytest
+            if: ${{ matrix.os != 'macos-latest'}}
             run: |
                 python -m pytest --cov --cov-append --cov-report=term-missing --ignore=tests/integration/ -vv
         -   name: Upload coverage
+            if: ${{ matrix.os != 'macos-latest' }}
```
LuckyMD, line 64: Without this, will the test workflow just fail on mac then? I guess then it's only designed for GA and not to test on personal laptop installs, no? Why remove this? Because it's difficult on an M1?

LuckyMD (approved):
This is fine if tests run through (also on mac?). Just 1 question on why mac support is removed.

setup.cfg
mumichae: The python-igraph package page indicates that it is deprecated and igraph should be used instead. But I think leidenalg also forces same igraph adn python-igraph versions

.github/workflows/test.yml
mumichae: I was getting segmentation faults on GA that I didn't get  on my personal computer (M1 macbook). I also tried to rerun an already passing run on a different branch and that was suddenly throwing errors. It could be related to some updated dependency, but it's hard to fix the problem if it is machine specific.

LuckyMD: Am I reading this correctly that the tests are only run on non-mac systems via GA though? Your experience suggests it works only on macOS

setup.cfg
LuckyMD: okay, thx.

.github/workflows/test.yml
mumichae: The tests were run on both Ubuntu and macOS, but only failed for macOS on GA. I couldn't reproduce the error on my personal macbook though and didn't want that issue to stall development

PR #339, opened 2022-10-10
Make rpy2 and anndata2ri optional

mumichae:


scib/metrics/lisi.py
```diff
@@ -188,6 +180,13 @@ def lisi_graph_py(
     Compute LISI score on shortes path based on kNN graph provided in the adata object.
     By default, perplexity is chosen as 1/3 * number of nearest neighbours in the knn-graph.
     """
+    try:
+        import rpy2.rinterface_lib.callbacks
+        import rpy2.rinterface_lib.embedded
+
+        rpy2.rinterface_lib.callbacks.logger.setLevel(logging.ERROR)
+    except ModuleNotFoundError as e:
+        raise OptionalDependencyNotInstalled(e)
```
LuckyMD, line 189: I don't think this is needed here. the only rpy2 command is an RRuntimeError, which is actually not needed anymore as it's all computed in python and C++, no? Need to check with @mbuttner, but this looks like legacy rpy2 code to me and so rpy2 isn't needed for lisi computation at all.

LuckyMD (changes requested):
Two things:
1. I think rpy2 isn't actually needed for lisi as we only use python and C++ and the only rpy2 parts are legacy code I think (in the lisi_graph function that is actually used and not deprecated) -> check this with @mbuttner 
2. If we make rpy2 optional, we also need to adapt the scib-pipeline install commands.

scib/metrics/lisi.py
mumichae: True, we wouldn't expect an R Runtime error. Do we still want to keep the behaviour of LISI returning NaN when the kNN graph fails?

mbuttner: LISI does not need any rpy2 anymore. As @LuckyMD said, it is all Python and C++. The old LISI implementation needed R.

LuckyMD: thanks for confirmation, @mbuttner ! And @mumichae: yes, NaN when kNN fails is good.

mumichae: OK, I'm catching a `RuntimeError` now instead

LuckyMD: Cool!

LuckyMD:
Before mergeable then only need:
 - [] test with different env setups
 - [] scib-pipeline PR for install scripts

mumichae:
For scib-pipeline, I don't think we need to change anything, since rpy2 is added as a dependency in the conda environments, regardless of the scib dependencies

mumichae:
scib-pipeline CI is running through using the dependencies https://github.com/theislab/scib-pipeline/pull/45

scib/preprocessing.py
```diff
@@ -774,7 +788,9 @@ def read_conos(inPath, dir_path=None):
     """
     from shutil import rmtree
 
-    import pandas as pd
+    import rpy2.rinterface_lib.callbacks
+    import rpy2.rinterface_lib.embedded
+    import rpy2.robjects as ro
```
LuckyMD, line 793: why are the rpy2 import statements not wrapped in a try-catch block in this file?

LuckyMD (approved):
looks good to me, but just 1 question about wrapping rpy2 imports in the preprocessing file.

PR #340, opened 2022-10-13
Update louvain call to a more recent method

johnarevalo:
Current version raises `BaseException: Could not construct partition: vector` error when calculating metrics that use Louvain clustering algorithm. As discussed in https://github.com/vtraag/louvain-igraph/issues/61 and [scanpy docs](https://scanpy-tutorials.readthedocs.io/en/latest/paga-paul15.html#Clustering-and-PAGA), the code should use `sc.tl.leiden` instead.

### DIFF
--- a/scib/metrics/clustering.py
+++ b/scib/metrics/clustering.py
@@ -79,7 +79,7 @@ def opt_louvain(
         sc.pp.neighbors(adata, use_rep=use_rep)
 
     for res in resolutions:
-        sc.tl.louvain(adata, resolution=res, key_added=cluster_key)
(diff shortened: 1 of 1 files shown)


LuckyMD:
Thanks for this PR @johnarevalo. I agree this is a good idea. The tests would also need to be adapted though. Is this the only place we have louvain in the code @mumichae?

mumichae:
Thanks for the PR. Leiden clustering was on the ToDo, but I didn't quite get to implement it. I wanted to implement a more flexible function that can use different clustering functions. I address this in #343

PR #343, opened 2022-10-24
General optimised clustering function

mumichae:
Create a new function for clustering for optimal resolution. Allows the user to specify whether they want to use the louvain, leiden or any other algorithm that works with an anndata and a cluster resolution.

Closes #340

### DIFF
--- a/tests/preprocessing/test_clustering.py
+++ b/tests/preprocessing/test_clustering.py
@@ -1,4 +1,5 @@
 import pandas as pd
+import scanpy as sc
 
 import scib
 from tests.common import LOGGER, assert_near_exact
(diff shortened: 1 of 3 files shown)


tests/preprocessing/test_clustering.py
```diff
@@ -18,3 +19,34 @@ def test_opt_louvain(adata_neighbors):
...
     assert res_max == 0.7
     assert_near_exact(score_max, 0.7432787576640969, diff=1e-3)
+
+
+def test_cluster_optimal_resolution_louvain(adata_neighbors):
+    res_max, score_max, score_all = scib.cl.cluster_optimal_resolution(
+        adata_neighbors,
+        label_key="celltype",
+        cluster_key="cluster",
+        cluster_function=sc.tl.louvain,
```
LuckyMD, line 29: we should test with leiden now i think... as louvain is being deprecated as i understand.

LuckyMD (approved):
you don't really need my code review for these types of changes...

PR #346, opened 2022-11-01
Update louvain to 0.8

scottgigante-immunai:
A new version of louvain has been released after quite some time. This PR updates scIB to use the new version, unpinning igraph.

PR #347, opened 2022-11-03
Fix code coverage

mumichae:
Switch to Codecov github actions as suggested by @scottgigante-immunai in #346


Zethson (approved):
Looks pretty good! Also how I'm doing it more or less. 


Well done. Hope it works :)

PR #350, opened 2022-11-21
Fix kBET import issue

### LINKED ISSUE #270: Proper kBet test
Currently, kBET is only tested for returning NaN as a result of failing to load the package (or other errors).

In order to fix this, we need

- [ ] environment with R and python (e.g. micromamba env from opsca-viash)
- [ ] new workflow in github actions (or udate the old one)
- [ ] update test

### LINKED ISSUE #349: kBET module not found error is supplanted by an np.nan return
This needs to be adapted to the same code as below @mumichae 

https://github.com/theislab/scib/blob/77ab015254754baa5ca3380bd592bcc9207241de/scib/metrics/kbet.py#L53-L56

mumichae:
Closes #349 
Closes #270

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ jobs:
             run: |
                 pre-commit run --color=always --show-diff-on-failure --all-files
 
-    test:
+    metrics:
         runs-on: ${{ matrix.os }}
         strategy:
             matrix:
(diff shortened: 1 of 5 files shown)


lazappi:
It's probably better to use **{remotes}** to install stuff rather than **{devtools}**. It has much fewer dependencies so it will make things faster.

LuckyMD (approved):
looks good from my side. But maybe you want to take on board @lazappi's comment above.

mumichae:
> It's probably better to use **{remotes}** to install stuff rather than **{devtools}**. It has much fewer dependencies so it will make things faster.

Thanks, that'll make things much easier!

PR #361, opened 2023-01-06
Make louvain an optional dependency

### LINKED ISSUE #360: louvain deprecation
Hi,
I'm trying to install this software as part of the [EasyBuild](https://easybuilders.github.io/)  package manager. For that, I need to install it from source.
However, my issue is that this package https://pypi.org/project/louvain/ is apparently deprecated and has probably been replaced by https://pypi.org/project/leidenalg/ .
I cannot even access documentation of `louvain`...
So my question is - is the `louvain` package still needed and used or can I simply patch the `setup.cfg` file and ignore it?
Thanks!

mumichae:
Addresses and closes #360

### DIFF
--- a/tests/integration/test_harmony.py
+++ b/tests/integration/test_harmony.py
@@ -10,13 +10,9 @@ def test_harmony(adata_paul15_template):
     )
 
     # check NMI after clustering
-    res_max, score_max, _ = scib.cl.opt_louvain(
-        adata,
-        label_key="celltype",
-        cluster_key="cluster",
-        plot=False,
-        inplace=True,
+    res_max, score_max, _ = scib.cl.cluster_optimal_resolution(
+        adata, label_key="celltype", cluster_key="cluster", return_all=True
     )
     LOGGER.info(f"max resolution: {res_max}, max score: {score_max}")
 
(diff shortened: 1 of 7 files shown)


tests/integration/test_harmony.py
```diff
@@ -10,13 +10,9 @@ def test_harmony(adata_paul15_template):
...
 
     # check NMI after clustering
-    res_max, score_max, _ = scib.cl.opt_louvain(
-        adata,
-        label_key="celltype",
-        cluster_key="cluster",
-        plot=False,
-        inplace=True,
+    res_max, score_max, _ = scib.cl.cluster_optimal_resolution(
+        adata, label_key="celltype", cluster_key="cluster", return_all=True
```
LuckyMD, line 14: order of parameters is different from order in `conftest.py` could you make this consistent?

LuckyMD (approved):
Just one minor change in order of parameters. Otherwise, is leiden already the default?

tests/integration/test_harmony.py
mumichae: yes, leiden is the default for the `cluster_optimal_resolution` function.

LuckyMD (approved):

PR #362, opened 2023-01-06
Add perfect isolabel ASW test

### LINKED ISSUE #359: Add perfect score metric tests for isolated labels
We are having some issues in open problems with the iso label ASW metric not giving perfect scores when one-hot-encoded cell type labels are used as embedding (with some jitter). We should add perfect score tests for this metric (and maybe more?) in the repo.

See https://github.com/openproblems-bio/openproblems/issues/685

Would you be able to quickly add this @mumichae ?

mumichae:
closes #359

### DIFF
--- a/tests/metrics/test_isolated_label.py
+++ b/tests/metrics/test_isolated_label.py
@@ -1,5 +1,25 @@
+import numpy as np
+
 import scib
 from tests.common import LOGGER, assert_near_exact
+
+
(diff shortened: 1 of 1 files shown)


LuckyMD:
This works here?... very strange as it doesn't in open problems... but thanks!

LuckyMD (approved):

Issue #367, opened 2023-01-26
isolated label scores do not return expected results


danielStrobl:
https://github.com/theislab/scib/blob/82444efd3e4378c487001e701a4a1209ef954aaa/scib/metrics/isolated_labels.py#L229-L230

This function gives the mean over all cells and not only the mean for the cells that are part of the isolated label.
This leads to much lower scores with the baselines as observed here: https://github.com/openproblems-bio/openproblems/issues/685

mumichae:
should be solved by #370

PR #368, opened 2023-01-27
Update isolated_labels.py

danielStrobl:
As mentioned in https://github.com/theislab/scib/issues/367, the isolated label metrics do not compute what they should.
By calculating the individual silhouette scores and then taking the mean over only the cells of the cluster, this should be fixed.

### DIFF
--- a/scib/metrics/isolated_labels.py
+++ b/scib/metrics/isolated_labels.py
@@ -226,7 +226,7 @@ def score_isolated_label(
     else:
         # AWS score between isolated label vs rest
         adata.obs[iso_label_key] = adata.obs[label_key] == isolated_label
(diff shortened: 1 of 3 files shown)


LuckyMD:
I guess you'll need to adapt the tests as well for this.

LuckyMD (approved):
Conceptually this looks correct, if silhouette_samples gives you the silhouette score per cell. See comment above about adding tests.

LuckyMD:
Please also adapt the perfect score test

danielStrobl:
ah bit too fast

LuckyMD:
@mumichae could you still review this please?

PR #369, opened 2023-01-27
Update test_isolated_label.py

danielStrobl:
add iso label asw perfect baseline


LuckyMD (approved):

danielStrobl:
hmm, not sure why the tests fail here, if I run this in a jupyter notebook I get a value close to 1

mumichae:
@danielStrobl Do we still need this PR? The changes should be added in #370

PR #370, opened 2023-01-30
Run iso label as discussed on friday

danielStrobl:
Iso label ASW as discussed on friday. Tested locally in jupyter notebook returning 1 for baseline dataset.

### DIFF
--- a/scib/metrics/isolated_labels.py
+++ b/scib/metrics/isolated_labels.py
@@ -225,7 +225,7 @@ def score_isolated_label(
         score = max_f1(adata, label_key, iso_label_key, isolated_label, argmax=False)
     else:
         # AWS score between isolated label vs rest
-        
+
         if "silhouette_temp" not in adata.obs:
             adata.obs["silhouette_temp"] = silhouette_samples(
                 adata.obsm[embed], adata.obs[label_key]
--- a/scib/metrics/isolated_labels.py
+++ b/scib/metrics/isolated_labels.py
@@ -225,10 +225,12 @@ def score_isolated_label(
         score = max_f1(adata, label_key, iso_label_key, isolated_label, argmax=False)
     else:
         # AWS score between isolated label vs rest
+        
+        if 'silhouette_temp' not in adata.obs:
+            adata.obs["silhouette_temp"] = silhouette_samples(
+                adata.obsm[embed], adata.obs[label_key]
+            )
         adata.obs[iso_label_key] = adata.obs[label_key] == isolated_label
-        adata.obs["silhouette_temp"] = silhouette_samples(
-            adata.obsm[embed], adata.obs[label_key]
-        )
         score = adata.obs[adata.obs[iso_label_key]].silhouette_temp.mean()
 
     if verbose:
--- a/scib/metrics/isolated_labels.py
+++ b/scib/metrics/isolated_labels.py
@@ -158,9 +158,9 @@ def isolated_labels(
     # 2. compute isolated label score for each isolated label
     scores = {}
     if not cluster:
-            adata.obs["silhouette_temp"] = silhouette_samples(
-                adata.obsm[embed], adata.obs[label_key]
-            )
+        adata.obs["silhouette_temp"] = silhouette_samples(
+            adata.obsm[embed], adata.obs[label_key]
+        )
     for label in isolated_labels:
         score = score_isolated_label(
             adata,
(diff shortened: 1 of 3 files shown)


LuckyMD (changes requested):

scib/metrics/isolated_labels.py
```diff
@@ -227,7 +227,7 @@ def max_f1(adata, label_key, cluster_key, label, argmax=False):
         # AWS score between isolated label vs rest
         adata.obs[iso_label_key] = adata.obs[label_key] == isolated_label
```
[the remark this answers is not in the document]
danielStrobl: kind of, this is still used to filter the silhouette scores to only the ones from the isolated label.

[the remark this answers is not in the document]
danielStrobl: could of course do this differently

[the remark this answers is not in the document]
mumichae: for readability might be better above the aggregation in line 232

tests/metrics/test_isolated_label.py
```diff
@@ -57,3 +57,16 @@ def test_isolated_labels_perfect(adata_pca):
...
+    adata_pca.obsm["X_emb"] = _random_embedding(partition=adata_pca.obs["celltype"])
+    score = scib.me.isolated_labels_asw(
+        adata_pca,
+        label_key="celltype",
+        batch_key="batch",
+        embed="X_emb",
+        verbose=True,
+    )
+    LOGGER.info(f"score: {score}")
+    assert_near_exact(score, 0.9983715917167036, diff=1e-12)
```
mumichae, line 72: must have missed adding a test for the ASW isolabel score 🙈 Thanks for adding that.

Might be better to compare with reference score 1 and adjust the tolerance accordingly (since this is supposed to represent a perfect score)

scib/metrics/isolated_labels.py
```diff
@@ -227,7 +227,7 @@ def max_f1(adata, label_key, cluster_key, label, argmax=False):
...
         adata.obs["silhouette_temp"] = silhouette_samples(
-            adata.obsm[embed], adata.obs[iso_label_key]
+            adata.obsm[embed], adata.obs[label_key]
```
mumichae, line 230: Since the ASW is run on the complete data and would always return the same silhouette scores, it might make sense to compute the scores once for the complete dataset and then aggregate it per isolated label score. That would require passing those scores to `score_isolated_label` (or adding them e.g. to `adata.uns` so they can be accessed within `score_isolated_label`), but I'd argue it's worth it, particularly for larger datasets.

mumichae (changes requested):
Looking good overall, would just request some changes for computational efficiency, but happy to address them in a different PR if necessary

scib/metrics/isolated_labels.py
danielStrobl: should work by just checking if the column exists, no?

danielStrobl: this could however be tricked by just adding a column silhouette_temp=1

mumichae: true, should be per obs. But I think it would be better to explicitly pass it to the function, just to be sure

scib/metrics/isolated_labels.py
```diff
@@ -105,13 +105,18 @@ def isolated_labels_asw(
...
+        adata.obsm[embed], adata.obs[label_key]
+    )
     return isolated_labels(
         adata,
         label_key=label_key,
         batch_key=batch_key,
         embed=embed,
         cluster=False,
         iso_threshold=iso_threshold,
+        silhouette_pre=True,
```
mumichae, line 119: Would be good to have silhouette precomputation in the `score_isolated_label` function only

scib/metrics/isolated_labels.py
```diff
@@ -158,7 +164,13 @@ def isolated_labels(
     scores = {}
     for label in isolated_labels:
```
mumichae, line 165: Ideally have checking whether ASW or F1 is used here and the precompute the ASW

scib/metrics/isolated_labels.py
```diff
@@ -124,6 +129,7 @@ def isolated_labels(
     cluster=True,
     iso_threshold=None,
     return_all=False,
+    silhouette_pre=False,
```
mumichae, line 132: remove this parameter

mumichae (changes requested):

mumichae (approved):

PR #371, opened 2023-02-01
Scale isolated label ASW

mumichae:
Isolated label ASW is not scaled, so that the expected output range would be between -1 and +1

https://github.com/theislab/scib/blob/98ee9a20260993ca5df960054dcc82ff8fd75b20/scib/metrics/isolated_labels.py#L237-L249

**Solution:** Add scaling option to metric

### DIFF
--- a/scib/metrics/isolated_labels.py
+++ b/scib/metrics/isolated_labels.py
@@ -67,6 +67,7 @@ def isolated_labels_asw(
     batch_key,
     embed,
     iso_threshold=None,
+    scale=True,
     verbose=True,
 ):
     """Isolated label score ASW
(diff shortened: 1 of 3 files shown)


danielStrobl (approved):
Looks good! Was wondering about that `test_isolated_labels_silhouette` one, it's a duplicate from `test_isolated_label_metrics.py`, right?

LuckyMD (approved):
Looks good to me. The same scaling as in cell type ASW, and it's scaled per label, so each label has equal weight.

LuckyMD:
We'll need a new release and then an open problems PR to pin to that release @danielStrobl

PR #372, opened 2023-02-01
Add GPU_id param for DESC

johnarevalo:
Without the `GPU_id` param, DESC method does not use the GPU.


LuckyMD (approved):
Looks good to me. Thanks for the PR! We didn't actually test any of the methods on GPU, but this doesn't break the CPU workflow, so all good.

PR #390, opened 2023-11-29
fix: check file empty

chyroc:


chyroc:
@mumichae Sorry to bother you, could you please review this PR for me? If there are any problems, I will correct them immediately.

PR #394, opened 2024-01-05
Skip labels before loop

johnarevalo:
When there are many labels to be skipped, having this validation inside the `for` loop makes kbet computation significantly slow:

https://github.com/theislab/scib/blob/ed3e2846414ca1e3dc07552c0eef1e68d82230d4/scib/metrics/kbet.py#L114-L122

This PR extract such checks outside the `for` loop to do them in a single pass.


mumichae (approved):

adamgayoso:
@johnarevalo would you be willing to make a similar PR to our scib-metrics package here? https://github.com/YosefLab/scib-metrics

johnarevalo:
Sure! I'll port these changes to scib-metrics and tag you when the PR is ready.

PR #395, opened 2024-01-09
Refactor LISI metrics

mumichae:
- [x] remove computational bottlenecks
- [ ] ~start from harmonypy implementation for LISI~ TODO in new PR


scib/metrics/lisi.py
```diff
@@ -240,8 +237,9 @@ def lisi_graph_py(
     if verbose:
         print("Convert nearest neighbor matrix and distances for LISI.")
 
-    batch = adata.obs[obs_key].cat.codes.values
-    n_batches = len(np.unique(adata.obs[obs_key]))
+    adata.obs[obs_key] = adata.obs[obs_key].astype("category")
+    batch_labels = adata.obs[obs_key].cat.codes.values
+    n_batches = len(batch_labels)
```
VladimirShitov, line 242: This will give you not the number of batches, but the number of observations (i.e., `adata.n_obs`) in the updated code.

The most compact way to get the number of batches would be
`n_batches = adata.obs[obs_key].nunique()`

VladimirShitov (changes requested):
I believe, there is a bug with `n_batches` atm, could you please check it? Otherwise, looks good :)

PR #408, opened 2024-04-22
Optimise PC regression

mumichae:


scib/metrics/pcr.py
```diff
@@ -216,18 +271,39 @@ def pc_regression(
...
+    if verbose:
+        print(f"Use {n_threads} threads for regression...")
+    if n_threads == 1:
+        r2 = []
+        for i in tqdm(range(n_comps), total=n_comps):
+            r2_score = linreg_method(X=covariate, y=X_pca[:, [i]])
+            r2.append(np.maximum(0, r2_score))
+    else:
+        with ThreadPoolExecutor(max_workers=n_threads) as executor:
+            run_r2 = executor.map(linreg_method, [covariate] * n_comps, X_pca.T)
```
VladimirShitov, line 286: What is parallelized here? And why is `X_pca` transposed when there are multiple threads but not the single one? Having some comments in the code would be nice :)

scib/metrics/pcr.py
```diff
@@ -216,18 +271,39 @@ def pc_regression(
...
     Var = pca_var / sum(pca_var) * 100
     R2Var = sum(r2 * Var) / 100
 
     return R2Var
+
+
+def linreg_sklearn(X, y):
+    from sklearn.linear_model import LinearRegression
+
+    lm = LinearRegression()
```
VladimirShitov, line 298: There is an `n_jobs` parameter with a default 1. Setting it to the number of threads would allow faster calculation with sklearn only

scib/metrics/pcr.py
```diff
@@ -216,18 +271,39 @@ def pc_regression(
...
+    lm.fit(X, y)
+    r2_score = lm.score(X, y)
+    np.maximum(0, r2_score)
+    return r2_score
+
+
+def linreg_np(X, y):
```
VladimirShitov, line 305: I wonder if for PC regression we can do it even more efficiently. According to chatGPT, both `sklearn.linear_model.LinearRegression` and `numpy.lstsq` apply SVD to `X`. PCA, however, does the same (and additionally subsetting the components), so we don't need to recalculate SVD twice. Check this chat for more info (I am not sure that the information from it is 100% correct though): https://chat.openai.com/share/f14c210f-6620-4fba-9047-57fdccb55955

tests/metrics/test_pcr_metrics.py
```diff
@@ -42,5 +79,5 @@ def test_pcr_batch_embedding(adata):
         n_comps=50,
         scale=True,
     )
-    LOGGER.info(f"using embedding: {score}")
+    LOGGER.info(f"using X_emb: {score}")
```
VladimirShitov, line 82: Why do you use proper logging in tests but simple prints in the code? Would be nice to use loggers everywhere :)

VladimirShitov (approved):
Overall, looks super nice. Sometimes, the code is unclear. I left some comments on potential improvements :)

scib/metrics/pcr.py
mumichae: I'm parallelising over all the PC components and X_pca needs to be transposed so that the matrix columns are passed (not the rows). But good point, will add a comment

scib/metrics/pcr.py
mumichae: You mean for the sklearn function? I would avoid threading within an already threaded loop, since that would square the number of threads requested in the worst case

scib/metrics/pcr.py
mumichae: I see, so you mean doing an SVD/PCA once and then computing the pseudo-inverse for each covariate? That's a cool idea! However, I think I'll keep it as is (i.e. minimal coding) for now and revisit it in another PR.

tests/metrics/test_pcr_metrics.py
mumichae: Good question, the testing code was obviously written way later than the actual function code. This would require refactoring all functions. I'd move that to when I have more time (since I feel like that would be way lower prio)

scib/metrics/pcr.py
VladimirShitov: Exactly! Yeah, this will likely not save you too much time, but it would make the code more complex. Maybe a cool thing to try later with proper testing that it works correctly

VladimirShitov (approved):
Looks great to me :)

VladimirShitov (approved):
Looks good in general :) I left a few questions

scib/metrics/pcr.py
```diff
@@ -280,14 +278,23 @@ def pc_regression(
...
-    R2Var = sum(r2 * Var) / 100
+        r2 = []
+        # parallelise over all principal components
+        with tqdm(total=n_comps) as pbar:
+            with ThreadPoolExecutor(max_workers=n_threads) as executor:
+                futures = [
+                    executor.submit(linreg_method, X=covariate, y=pc) for pc in X_pca.T
+                ]
+                for future in as_completed(futures):
+                    r2.append(future.result())
```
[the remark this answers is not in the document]
mumichae: Good point. I removed the multi-threading completely, because it also didn't seem to speed up the process much and just ate up memory.

scib/metrics/pcr.py
```diff
@@ -299,11 +306,14 @@ def linreg_sklearn(X, y):
...
 
 def linreg_np(X, y):
-    coefficients, residuals, _, _ = np.linalg.lstsq(X, y, rcond=None)
+    _, residuals, _, _ = np.linalg.lstsq(X, y, rcond=None)
+    if residuals.size == 0:
+        residuals = np.array([0])
+    rss = residuals[0] if len(residuals) > 0 else 0
     tss = np.sum((y - y.mean()) ** 2)
-    r2_score = 1 - (residuals[0] / tss)
+    r2_score = 1 - (rss / tss)
```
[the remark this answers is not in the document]
mumichae: I guess you could use the sklearn function for this, but then you'd need to import it. I'll leave it as-is for now

scib/metrics/cell_cycle.py
```diff
@@ -190,16 +198,28 @@ def get_pcr_before_after(
     covariate = raw_sub.obs[["S_score", "G2M_score"]]
 
     # PCR on adata before integration
-    if recompute_pcr:
+    if recompute_pcr:  # TODO: does this work for precomputed values?
```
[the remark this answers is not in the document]
mumichae: It doesn't, but at the same time you want to do PCA on subsets of the data, which the user most likely won't have precomputed anyway.
```suggestion
    if recompute_pcr:
```

mumichae (commented):
Thanks, I addressed your comments, but only changed the linear regression part to use multiple regression. According to my tests the output is sufficiently similar

PR #410, opened 2024-07-05
M.A might not be available, you should use M.toarray()

Zhongzheng99:
In some versions of scipy(e.g. 1.14), M.A might not be available. In such cases, you should use M.toarray() to convert the sparse matrix to a dense one.


Zhongzheng99:
#412

mumichae (approved):
Closes https://github.com/theislab/scib/issues/412

PR #421, opened 2025-01-10
Extend cell cycle gene scoring

mumichae:
* add new organism gene sets
* smarter gene ID/ gene name usage


scib/preprocessing.py
```diff
@@ -653,43 +656,87 @@ def score_cell_cycle(adata, organism="mouse"):
...
+        n_genes = 0
+        for col in columns:
+            _genes = [g for g in df[col] if g in adata.var_names]
+            if len(_genes) > n_genes:  # pick largest overlapping set
+                n_genes = len(_genes)
+                genes = _genes
+
+        if n_genes == 0:
+            rand_choice = np.random.randint(1, adata.n_vars, 10)
+            rand_genes = adata.var_names[rand_choice].tolist()
```
VladimirShitov, line 683: Simply `np.random.choice(adata.var_names, 10)` would work instead of these 2 lines :)

VladimirShitov: Also, I'd add a comment on why you are doing it. I got it after a while, but it was not obvious right away

scib/preprocessing.py
```diff
@@ -653,43 +656,87 @@ def score_cell_cycle(adata, organism="mouse"):
...
+
+    # filter gene sets across data
+    s_genes = filter_genes(adata, gene_map.query("phase == 'S'"))
+    g2m_genes = filter_genes(adata, gene_map.query("phase == 'G2/M'"))
 
+    # compute scores
     sc.tl.score_genes_cell_cycle(adata, s_genes=s_genes, g2m_genes=g2m_genes)
 
 
+def get_cell_cycle_genes(organism: str):
```
VladimirShitov, line 700: Would be great to have a list of available organisms as a data type instead of just `str`. There is some mapping of possible options, but there can be much more like "C.elegans" or "Human", for example. Having an explicit list would allow one to simply copy and paste from it

scib/preprocessing.py
```diff
@@ -653,43 +656,87 @@ def score_cell_cycle(adata, organism="mouse"):
...
+        "mus musculus": "mus_musculus",
+        "human": "homo_sapiens",
+        "homo sapiens": "homo_sapiens",
+        "c_elegans": "caenorhabditis_elegans",
+        "caenorhabditis elegans": "caenorhabditis_elegans",
+        "c elegans": "caenorhabditis_elegans",
+        "zebrafish": "danio_rerio",
+        "danio rerio": "danio_rerio",
+    }
+    organism_map |= {x: x for x in organism_map.values()}
```
VladimirShitov, line 729: Could you add a comment on what's happening here? Also, perhaps a more readable version would be
```suggestion
    organism_map.update({x: x for x in organism_map.values()})
```

scib/preprocessing.py
```diff
@@ -653,43 +656,87 @@ def score_cell_cycle(adata, organism="mouse"):
...
+
+    assert organism in organism_map, f"organism {organism} not supported"
```
VladimirShitov, line 731: You can output the list of supported organism names as well.
```suggestion
    assert organism in organism_map, f"organism {organism} not supported. Supported organisms: {list(organism_map.values()})"
```

scib/resources/convert_genes.py
```diff
@@ -0,0 +1,112 @@
+import requests
+from tqdm import tqdm
+
+
+def get_gene_name_from_ensembl(gene_ids: list, species: str):
```
VladimirShitov, line 5: Isn't there a simpler way to do it? E.g. using `bionty` Gene [lookup](https://docs.lamin.ai/bionty.gene#bionty.Gene).  It probably does the same under the hood, but if something breaks in API, it is their responsibility to fix it, and not your

VladimirShitov: It can be fine to leave it like that of course, if you don't want to add an additional dependency

tests/conftest.py
```diff
@@ -97,3 +97,36 @@ def adata_clustered(adata_neighbors):
...
+
+    import requests
+
+    dataset_name = request.param
+    url = DATASETS[dataset_name]
+
+    response = requests.get(url)
+    response.raise_for_status()
+
+    adata = sc.read_h5ad(BytesIO(response.content))
```
VladimirShitov, line 121: scanpy can read URLs natively:
```suggestion
    adata = sc.read("placeholder_path.h5ad", backup_url=url)
```

The file will be saved to `placeholder_path` this way, so maybe worth creating a temp file

tests/preprocessing/test_gene_scoring.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+import scanpy as sc
+
+import scib
+
+
+def test_mouse(adata_paul15):
+    scib.pp.score_cell_cycle(
```
VladimirShitov, line 8: Could you add some asserts to the test function itself? E.g. that columns `["S_score", "G2M_score", "phase"]` were added to the adata

VladimirShitov (changes requested):
In general, looks great! I added a few minor comments to improve readability. Would be great to resolve them before merging

scib/preprocessing.py
mumichae: Thanks, that was an older piece of code that I hadn't considered improving

scib/preprocessing.py
mumichae: You mean using e.g. Literal?

scib/resources/convert_genes.py
mumichae: Good point! With bionty it would probably be more elegant, but I didn't want to include any additional dependencies here and this script is only for reproducibility purposes (i.e. will never be run in the package)

tests/conftest.py
mumichae: Thanks, I totally forgot about that feature!

VladimirShitov (approved):
Looks great!

